PR #1, opened 2019-06-25
Added util file and the connect method used in the creator

fatcat2:
## Contents:
* `connect`: returns a DB API connection
* `generate_keys`: returns a PEM encoded RSA public/private keypair.
* `write_to_file`: writes the server certificate authority, RSA public key, and ephemeral certificate to `.pem` files.


utils.py
```diff
@@ -0,0 +1,68 @@
...
+    public_key = private_key_obj.public_key().public_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PublicFormat.SubjectPublicKeyInfo
+    )
+    return private_key, public_key
+
+
+# Helper function to write the serverCaCert, ephemeral certificate and private
+# key to .pem files
+def write_to_file(serverCaCert, ephemeralCert, priv_key):
```
kurtisvg: Do these have to be in files? Can you use in memory ones instead?

requirements.txt
```diff
@@ -0,0 +1,17 @@
+asn1crypto==0.24.0
```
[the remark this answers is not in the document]
fatcat2: These are from a `pip freeze` command and are also installed when you run `pip install -r requirements.txt` in some of the Python sample docs. The Python sample docs only list the main packages, so I'll match what they do.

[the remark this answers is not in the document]
kurtisvg: Typically you only want to include the direct dependency and not the ones it relies on. This way if they bump there versions, you just need to update the direct dependency and not the other options.

utils.py
fatcat2: I think they do, the pymysql connector only takes filepaths as arguments. I'll ask Chris Wilcox or Thea Flowers to see if Python has a way to fake it somehow.

fatcat2: Looks like they have to be in files.

PR #2, opened 2019-06-26
ephemeral_certificate function branch

fatcat2:
Adding in a function to connector.py that requests an ephemeral certificate from the Cloud SQL instance specified. The ephemeral certificate is used to create a SSL connection with the Cloud SQL Instance.


connector.py
```diff
@@ -0,0 +1,53 @@
...
+    # TODO(ryachen@) Add checks to ensure service object is valid.
+
+    request = service.sslCerts().createEphemeral(
+        project=project,
+        instance=instance,
+        body={
+            'public_key': pub_key
+        }
+    )
+    response = request.execute()
```
busunkim96: If this is a `requests` object you can do `.raise_for_status()` and an exception will be raised if there was an error. https://2.python-requests.org/en/master/api/#requests.Response.raise_for_status

connector.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    if (
+        service is None or
+        project is None or
+        instance is None or
+        pub_key is None
+    ):
+        print("Cannot take None as an argument.")
+        exit()
```
busunkim96: Could you raise an exception instead of calling `exit()`? That way someone using this library can catch the exception and do additional processing.

connector.py
```diff
@@ -0,0 +1,53 @@
...
+limitations under the License.
+"""
+
+
+def get_ephemeral(service, project, instance, pub_key):
+    """
+    A helper function that requests an ephemeral certificate from the
+    Cloud SQL Instance.
+
+    Takes in a service object, the project name, the instance name and
```
busunkim96: Would you mind putting this into the google format for docstrings? :) That will add info on the types of the arguments and should make it easier for other folks to use. You'll also be able to get nice generated docs.

Example: https://github.com/googleapis/google-cloud-python/blob/3d324389b92d43e52486f0fe2aca8b41e950640c/bigquery_storage/google/cloud/bigquery_storage_v1beta1/client.py#L85-L115
Documentation on google style docstrings: http://google.github.io/styleguide/pyguide.html#381-docstrings

test.py
```diff
@@ -1,32 +1,40 @@
...
 
 import pytest  # noqa: F401; pylint: disable=unused-variable
 from utils import generate_keys
 
 
-# Test to check if objects are being produced from the generate_keys()
-# function.
 def test_generate_keys_1():
+    """
+    # Test to check if objects are being produced from the generate_keys()
```
busunkim96: Is this a stray `#`?

fatcat2: Yup, thanks for catching that!

connector.py
fatcat2: Looks like it isn't a `requests` object, but `request.execute()` raises an error when there's a 400 status code. So I guess the check isn't necessary haha.

connector.py
fatcat2: Will do, thanks for the links!

fatcat2:
FYI: @kurtisvg @jkwlui

connector.py
```diff
@@ -0,0 +1,61 @@
...
+
+def get_ephemeral(service, project_name, instance_name, pub_key):
+    """
+    A helper function that requests an ephemeral certificate from the
+    Cloud SQL Instance.
+
+    Args:
+        service: A service object created from the Google Python API client
+            library. Must be using the SQL Admin API.
+        project (String): A string representing the name of the project.
```
busunkim96:
```suggestion
        project (str): A string representing the name of the project.
```
Same for the two below :)

connector.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+
+def get_ephemeral(service, project_name, instance_name, pub_key):
+    """
+    A helper function that requests an ephemeral certificate from the
+    Cloud SQL Instance.
+
+    Args:
+        service: A service object created from the Google Python API client
```
busunkim96:
```suggestion
        service (googleapiclient.discovery.Resource): A service object created from the Google Python API client
```

connector.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+
+    if (
+        service is None or
+        project_name is None or
+        instance_name is None or
+        pub_key is None
+    ):
+        print("Cannot take None as an argument.")
+        raise(TypeError)
```
busunkim96: I think you can take care of both lines at once by doing:

```suggestion
        raise TypeError("Cannot take None as an argument")
```

connector.py
```diff
@@ -0,0 +1,61 @@
...
+    request = service.sslCerts().createEphemeral(
+        project=project_name,
+        instance=instance_name,
+        body={
+            'public_key': pub_key
+        }
+    )
+    response = request.execute()
+
+    print(response)
```
busunkim96: Was this here for debugging?

.env
```diff
@@ -0,0 +1,7 @@
+export GOOGLE_APPLICATION_CREDENTIALS=".auth.json"
```
kurtisvg: this file probably doesn't need to be here either

fatcat2: Deleted!

fatcat2:
@busunkim96 I made the changes you requested and committed them but Github still says I haven't made the changes. Would you mind taking a look for me?

test.py
```diff
@@ -1,32 +1,38 @@
...
+limitations under the License.
+"""
 
 import pytest  # noqa: F401; pylint: disable=unused-variable
 from utils import generate_keys
 
 
-# Test to check if objects are being produced from the generate_keys()
-# function.
 def test_generate_keys_1():
```
crwilcox: In general we try to make the names descriptive as it helps when it fails to quickly figure out why the change may have caused the test to fail. Something like `test_generate_keys_returns_bytes` or something of that sort?

I realize this is a bit out of scope of the PR so you can just keep this in mind for your next change if you would like.

connector.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def get_ephemeral(service, project_name, instance_name, pub_key):
+    """
+    A helper function that requests an ephemeral certificate from the
+    Cloud SQL Instance.
+
+    Args:
+        service (googleapiclient.discovery.Resource): A service object created
+          from the Google Python API client library. Must be using the SQL
```
crwilcox: Perhaps we should mention the package, or have a link to the documentation? Often users are confused between the gRPC and the Rest/Discovery client libraries.

noxfile.py
```diff
@@ -0,0 +1,89 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018 Google LLC
```
busunkim96: nit: 2019, please. :)

fatcat2: Oops, good catch!

connector.py
busunkim96: @fatcat2 Would you mind linking to the repo: https://github.com/googleapis/google-api-python-client ?

I'm currently in the process of moving around documentation for that library. :)

noxfile.py
```diff
@@ -0,0 +1,88 @@
...
+
+from __future__ import absolute_import
+import os
+import shutil
+import sys
+import nox
+
+BLACK_PATHS = ["connector.py", "tests", "noxfile.py"]
+
+if os.path.exists("samples"):
```
crwilcox: if you plan to have samples you can leave this, it isn't harmful either way but it is cruft

noxfile.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+
+from __future__ import absolute_import
+import os
+import shutil
+import sys
+import nox
+
+BLACK_PATHS = ["connector.py", "tests", "noxfile.py"]
```
crwilcox: It may make sense to use patterns. For instance `*.py` to get all in root. If that isn't desirable though by name is good.

setup.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from setuptools import setup, find_packages
+
+setup(name="cloud-sql-python-connector", packages=find_packages())
```
crwilcox: for now this is likely fine, but we will want to flesh this out before launch. Also, probably 2019 for copyright

busunkim96:
FYI @fatcat2 You can make a [.flake8](https://github.com/googleapis/google-cloud-python/blob/a96fa648a28521906740c0d3efd401e13fbb670a/tasks/.flake8)  to tell flake to ignore some warnings/errors.

PR #4, opened 2019-07-01
Organizing cloud-sql-python-connector

fatcat2:
Goal of this PR is to organize the package by putting files in appropriate folders.

FYI: @kurtisvg, @jkwlui


tests/test_unit.py
```diff
@@ -14,7 +14,7 @@
 limitations under the License.
 """
 
-import utils
+from google.cloud.sql import utils
```
kurtisvg: use `google.cloud.sql.connector`.

PR #5, opened 2019-07-02
Adding get_metadata function

fatcat2:
Adding the `get_metadata` function into `connector.py`.

FYI: @kurtisvg, @jkwlui

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -16,6 +16,29 @@ limitations under the License.
 
 import googleapiclient
 
+def get_metadata(service, proj_name, inst_name):
+    """
+    A helper function that requests metadata from the Cloud SQL Instance
+    and returns a dictionary containing the IP address and certificate
+    authority of the Cloud SQL Instance.
+
+    Args:
+        service (googleapiclient.discovery.Resource): A service object created
+          from the Google Python API client library. Must be using the SQL
+          Admin API. For more info check out
+          https://github.com/googleapis/google-api-python-client.
+        project (str): A string representing the name of the project.
+        instance (str): A string representing the name of the instance.
+            Usually found in environment variable 'CLOUD_SQL_INSTANCE_NAME.'
+
+    Returns:
+        An ephemeral certificate from the Cloud SQL instance that allows
+        authorized connections to the instance.
+
+    Raises:
+        TypeError: If any of the arguments are not the specified type.
+    """
+
 
 def get_ephemeral(service, proj_name, inst_name, pub_key):
     """
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -20,7 +20,7 @@ import googleapiclient
 def get_metadata(service, proj_name, inst_name):
     """
     A helper function that requests metadata from the Cloud SQL Instance
-    and returns a dictionary containing the IP address and certificate
+    and returns a dictionary containing the IP addresses and certificate
     authority of the Cloud SQL Instance.
 
     Args:
@@ -33,8 +33,8 @@ def get_metadata(service, proj_name, inst_name):
             Usually found in environment variable 'CLOUD_SQL_INSTANCE_NAME.'
 
     Returns:
-        An ephemeral certificate from the Cloud SQL instance that allows
-        authorized connections to the instance.
+        Returns a dictionary containing a dictionary of all IP addresses and
+        their type and a string representing the certificate authority.
 
     Raises:
         TypeError: If any of the arguments are not the specified type.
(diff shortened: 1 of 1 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -51,16 +51,18 @@ def get_metadata(service, proj_name, inst_name):
...
-        ip = res["ipAddresses"][0]["ipAddress"]
+    # Extract server certificate authority
+    serverCaCert = res["serverCaCert"]["cert"]
 
-        serverCaCert = res["serverCaCert"]["cert"]
+    # Map IP addresses to type.
+    ip_map = {ip["type"]: ip["ipAddress"] for ip in res["ipAddresses"]}
 
-        metadata = {"ip": ip, "ca": serverCaCert}
+    metadata = {"ip": ip_map, "ca": serverCaCert}
```
kurtisvg: nit: ip -> ip_address
nit: ca -> server_ca_cert

google/cloud/sql/connector/connector.py
```diff
@@ -17,6 +17,54 @@
 import googleapiclient
 
 
+def get_metadata(service, proj_name, inst_name):
```
crwilcox: in general it is a good idea not to abbreviate. `project`, `instance` would also better match other interfaces we have: https://github.com/googleapis/google-cloud-python/blob/e6898be297e4938cce02c00d221eaf5d88002162/core/google/cloud/client.py#L199

google/cloud/sql/connector/connector.py
```diff
@@ -17,6 +17,54 @@
...
+    """
```
crwilcox: Sort of a higher level discussion the python team should look at. Our other modules are using sphinx style docstrings. There are two things I would want to verify if we used google style.

1) I have heard that napoleon and sphinx should be able to translate google style.
2) I would want to know that IDEs/plugins that do typehinting are also good at this. I think the sphinx style is verbose which helps out a bit. I don't use PyCharm but it seems they support sphinx style? https://www.jetbrains.com/help/pycharm/using-docstrings-to-specify-types.html

You are welcome to leave this and if at documentation generation time we struggle, we might have to address this. just heads up.

For an example of what I mean: https://github.com/googleapis/google-cloud-python/blob/master/storage/google/cloud/storage/client.py#L35

google/cloud/sql/connector/connector.py
```diff
@@ -17,6 +17,54 @@
...
+    Args:
+        service (googleapiclient.discovery.Resource): A service object created
+          from the Google Python API client library. Must be using the SQL
+          Admin API. For more info check out
+          https://github.com/googleapis/google-api-python-client.
+        proj_name (str): A string representing the name of the project.
+        inst_name(str): A string representing the name of the instance.
+            Usually found in environment variable 'CLOUD_SQL_INSTANCE_NAME.'
+
+    Returns:
```
crwilcox: Should specify the return type. which I think is `Dict[str, str]`?

google/cloud/sql/connector/connector.py
```diff
@@ -17,6 +17,54 @@
...
+    authority of the Cloud SQL Instance.
+
+    Args:
+        service (googleapiclient.discovery.Resource): A service object created
+          from the Google Python API client library. Must be using the SQL
+          Admin API. For more info check out
+          https://github.com/googleapis/google-api-python-client.
+        proj_name (str): A string representing the name of the project.
+        inst_name(str): A string representing the name of the instance.
+            Usually found in environment variable 'CLOUD_SQL_INSTANCE_NAME.'
```
crwilcox: if this is a common env var would it make sense for us to identify this and make this an optional param?

Similarly for project. The `ClientWithProject` class in `google-cloud-python/core` makes project optional and will identify the project automagically.

google/cloud/sql/connector/connector.py
```diff
@@ -17,6 +17,54 @@
...
+            "Arguments must be as follows: "
+            + "service (googleapiclient.discovery.Resource), "
+            + "proj_name (str) and inst_name (str)."
+        )
+
+    req = service.instances().get(project=proj_name, instance=inst_name)
+    res = req.execute()
+
+    # Extract server certificate authority
+    serverCaCert = res["serverCaCert"]["cert"]
```
crwilcox: is there a chance these properties aren't in the result?

fatcat2: Don't think so, I'm pretty sure it's required in the response.

Here's the link to the API: https://cloud.google.com/sql/docs/mysql/admin-api/v1beta4/instances

google/cloud/sql/connector/connector.py
kurtisvg: I missed this before - but generally they'll specify the instance name in the connect method (we probably don't want to advertise they should set it via this env var)

google/cloud/sql/connector/connector.py
fatcat2: I'll open a PR to change the documentation to Sphinx style

google/cloud/sql/connector/connector.py
```diff
@@ -17,7 +17,55 @@
...
+    Args:
+        service (googleapiclient.discovery.Resource): A service object created
+          from the Google Python API client library. Must be using the SQL
+          Admin API. For more info check out
+          https://github.com/googleapis/google-api-python-client.
+        proj_name (str): A string representing the name of the project.
+        inst_name(str): A string representing the name of the instance.
+
+    Returns:
+        Dict[str: Dict[str: str], str: str]
```
crwilcox: I can't map this to https://www.python.org/dev/peps/pep-0484/ and trying to figure out what this would be.

I think `so Dict[str, Union[Dict, str]]`?

google/cloud/sql/connector/connector.py
fatcat2: From previous experience, I'm pretty sure it would result in an earlier error. I'll add in a check just to be sure!

PR #6, opened 2019-07-08
Creating an interface for connect

fatcat2:
The purpose of this PR is to set a basic definition on what the `connect` method is supposed to do. Raises a `NotImplementedError` for now.

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -104,3 +104,34 @@ def get_ephemeral(service, project, instance, pub_key):
     response = request.execute()
 
     return response["cert"]
+
+
+def connect(instance_connection_string, connection_args=None):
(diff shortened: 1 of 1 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -104,3 +104,35 @@ def get_ephemeral(service, project, instance, pub_key):
     response = request.execute()
 
     return response["cert"]
+
+
+def connect(instance_connection_string, connection_args=None):
```
kurtisvg: Would `kwargs` be a better fit here?

fatcat2: I think `kwargs` would probably be a better fit, it allows for more flexibility on the developer's end. Also helps make it more clear what is being passed in.

PR #7, opened 2019-07-08
Create InstanceConnectionManager class

fatcat2:


google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,44 @@
...
+        # Validate connection string
+        pattern = "([\\S]+):([\\S]+):([\\S]+)"
+        match = re.match(pattern, instance_connection_string)
+
+        if match:
+            self.instance_connection_string = instance_connection_string
+            self.project = match[1]
+            self.instance = match[2]
+        else:
+            raise Exception(
```
kurtisvg: Should this be a generic exception, or a more specific exception so that users can catch it if it occurs?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,44 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import re
+
+
+class InstanceConnectionManager:
+    instance_connection_string = None
```
kurtisvg: Add type hints to these declarations.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,44 @@
...
+    loop = None
```
kurtisvg: Should these be private variables?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,44 @@
...
+    project = None
+    instance = None
+    credentials = None
+
+    def __init__(self, instance_connection_string, loop):
+        # Validate connection string
+        pattern = "([\\S]+):([\\S]+):([\\S]+)"
```
kurtisvg: Might be less complex to use split on `:` and validate there are 3 parts

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,44 @@
...
+        match = re.match(pattern, instance_connection_string)
+
+        if match:
+            self.instance_connection_string = instance_connection_string
+            self.project = match[1]
```
kurtisvg: What about region?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,44 @@
...
+            self.instance = match[2]
+        else:
+            raise Exception(
+                "Arg instance_connection_string must be in "
+                + "format: project:region:instance."
+            )
+
+        #
```
kurtisvg: nit: empty comment line?

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: Created `CloudSQLInstanceStringError` to resolve this.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,58 @@
...
+    _project: str = None
+    _region: str = None
+    _instance: str = None
+    _credentials: Credentials = None
+
+    def __init__(
+        self,
+        instance_connection_string: str,
+        loop: asyncio.unix_events._UnixSelectorEventLoop,
+    ) -> None:
```
crwilcox: is the intention to only support Python 3.5+? just want to verify.

fatcat2: Yup! Since Python 2.7 is set to be deprecated in five months.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,69 @@
...
+        # Validate connection string
+        connection_string_split = instance_connection_string.split(":")
+
+        if len(connection_string_split) == 3:
+            self.instance_connection_string = instance_connection_string
+            self.project = connection_string_split[1]
```
kurtisvg: connection_string_split is in the format `project:region:instance`

Also, region is not assigned.

requirements.txt
```diff
@@ -2,3 +2,4 @@ cryptography==2.7
 PyMySQL==0.9.3
 pytest==4.6.3
 google-api-python-client==1.7.9
+google-auth==1.6.3
```
kurtisvg: This is a requirement of `google-api-python-client` as well - do you need to list is separately or would it be better to just use it as a dependency of `google-api-python-client`?

tests/test_unit.py
```diff
@@ -35,3 +42,13 @@ def test_generate_keys_returns_bytes():
 
     res1, res2 = utils.generate_keys()
     assert isinstance(res1, bytes) and (isinstance(res2, bytes))
+
+
+def test_InstanceConnectionManager_connection_string_parsing():
```
kurtisvg: I would consider putting all of the `InstanceConnectionManager` tests into their own file. In general, I try to map the module layout to the testing layout. 

Also when testing, you should have one test for everything that goes right(unless there are multiple ways it can go right), and then additional tests for everything that could go wrong. 

So here you want something like `test_InstanceConnectionManager_init` and you test that it was created successfully. You also want something like `test_InstanceConnectionManager_wrong_connection_string` that verifies that the correct error is thrown when the argument in improperly formatted.

requirements.txt
fatcat2: I don't think I need to list this separately.

fatcat2:
#8: `InstanceConnectManager` now uses Sphinx-style documentation

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -0,0 +1,75 @@
...
+    _instance_connection_string: str = None
+    _project: str = None
+    _region: str = None
+    _instance: str = None
+    _credentials: Credentials = None
+
+    def __init__(
+        self,
+        instance_connection_string: str,
+        loop: asyncio.unix_events._UnixSelectorEventLoop,
```
kurtisvg: This type hint needs to be updated yet

tests/test_instanceconnectionmanager_unit.py
```diff
@@ -0,0 +1,48 @@
...
+from google.cloud.sql.connector.InstanceConnectionManager import (
+    InstanceConnectionManager,
+    CloudSQLConnectionError,
+)
+import asyncio
+
+
+def test_InstanceConnectionManager_connection_string():
+    """
+    Test to check whether the __init__() method of InstanceConnectionManager
```
kurtisvg: nit: Name should be more general

kurtisvg: super nit
```suggestion
    Test to check whether the __init__ method of InstanceConnectionManager
```

PR #9, opened 2019-07-09
Move get_metadata and get_certificate into InstanceConnectionManager

fatcat2:


google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -71,3 +72,91 @@ def __init__(
 
         # set current to future InstanceMetadata
         # set next to the future future InstanceMetadata
+
+    def _get_metadata(service, project, instance):
```
kurtisvg: nit: Probably would be good to type hint the function signatures

PR #13, opened 2019-07-12
Adding _perform_refresh to InstanceConnectionManager

fatcat2:
Adding a `_perform_refresh` function to `InstanceConnectionManager` to refresh the ephemeral certificate and instance metadata.

FYI: @kurtisvg, @jkwlui


tests/test_instanceconnectionmanager_unit.py
```diff
@@ -46,3 +47,14 @@ def test_InstanceConnectionManager_wrong_connection_string():
...
     with pytest.raises(CloudSQLConnectionError):
         InstanceConnectionManager("test-project:test-region", loop)
+
+
+def test_InstanceConnectionManager_perform_refresh():
+    """
+    Test to check whether _perform_refresh can run without any errors given a
+    live testing string.
+    """
+    test_connection_string = os.environ["CLOUD_SQL_CONNECTION_STRING"]
```
busunkim96: Is this something that you can assume is always in the environment?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -135,7 +135,7 @@ def _get_metadata(
 
         return metadata
 
-    def _get_ephemeral(service, project, instance, pub_key):
+    def _get_ephemeral(self, service, project, instance, pub_key):
```
kurtisvg: Should these have type hints?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -84,7 +84,7 @@ def __init__(
         # set next to the future future InstanceMetadata
 
     def _get_metadata(
-        service: Resource, project: str, instance: str
+        self, service: Resource, project: str, instance: str
```
kurtisvg: Should these have type hints?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -192,3 +192,25 @@ def _auth_init(self):
...
+        """Retrieves instance metadata and ephemeral certificate from the
+        Cloud SQL Instance.
+
+        :rtype: Dict[str, Union[Dict, str]]
+        :returns: A Dictionary containing the server's certificate authority,
+            a Dictionary containing the IP addresses of the instance, and
+            the latest ephemeral certificate.
+        """
+        metadata_result = self._get_metadata(
+            self._cloud_sql_service, self._project, self._instance
```
kurtisvg: Does `self._get_metadata` need to be an object method? If you are already passing in the required attributes, would it be better as static instead?

tests/test_instanceconnectionmanager_unit.py
kurtisvg: No, it's probably not - it would be a good idea to check if this is unset and either throw an error or skip the test (not sure which would be best for Python).

Also, please try to use `"INSTANCE_CONNECTION_NAME"` since this is how the SQL docs refer to it.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -85,6 +87,15 @@ def __init__(
         # set current to future InstanceMetadata
         # set next to the future future InstanceMetadata
 
+    def __del__(self):
+        async def close_client_session():
+            await self._client_session.close()
```
kurtisvg: Why does this need to be wrapped?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -85,6 +87,15 @@ def __init__(
         # set current to future InstanceMetadata
         # set next to the future future InstanceMetadata
 
+    def __del__(self):
```
kurtisvg: This function could use a brief comment explaining why this is necessary.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -230,3 +247,42 @@ def _auth_init(self):
...
+                self._project,
+                self._instance,
+                self._pub_key.decode("UTF-8"),
+            )
+        )
+
+        instance_data_task = self._loop.create_task(
+            self._get_context(metadata_future, ephemeral_future)
+        )
+        instance_data_task.add_done_callback(self._threadsafe_refresh)
```
kurtisvg: This isn't implemented yet?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -211,6 +222,12 @@ def __init__(
 
         return ret_dict["cert"]
 
+    async def _get_context(self, ephemeral_task, metadata_task):
```
kurtisvg: would prefer `create_context` rather than get

tests/test_instanceconnectionmanager_unit.py
```diff
@@ -121,3 +121,23 @@ def test_InstanceConnectionManager_get_metadata():
...
+        raise KeyError(
+            "Please set environment variable 'INSTANCE_CONNECTION"
+            + "_NAME' to a valid Cloud SQL connection string."
+        )
+
+    loop = asyncio.new_event_loop()
+    icm = InstanceConnectionManager(connect_string, loop)
+    fut = icm._perform_refresh()
+
+    assert isinstance(fut, asyncio.Task)
```
kurtisvg: It looks like this test only checks to make sure it returns something, not that it actually works?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -78,12 +95,41 @@ def __init__(
                 "Arg instance_connection_string must be in "
                 + "format: project:region:instance."
             )
+
         self._loop = loop
         self._auth_init()
         self._priv_key, self._pub_key = generate_keys()
```
kurtisvg: Can you open an issue for adding generate_keys asynchronously?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -78,12 +95,41 @@ def __init__(
...
+        self._mutex = threading.Lock()
+
+        async def create_client_session():
+            return aiohttp.ClientSession()
+
+        self._client_session = asyncio.run_coroutine_threadsafe(
+            create_client_session(), loop=self._loop
+        ).result()
+
+        print("instance data")
```
kurtisvg: Should this print statement be here? 

Might be worth considering using `logging` instead  so that we can leave these and use them to debug. Ask Chris how to determine which logger should be used.

kurtisvg: same for other print statements

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -204,14 +254,82 @@ def __init__(
...
+            )
+        )
+
+        ephemeral_task = self._loop.create_task(
+            self._get_ephemeral(
+                self._client_session,
+                self._credentials,
+                self._project,
+                self._instance,
+                self._pub_key.decode("UTF-8"),
```
kurtisvg: nit: If we decode this pub_key every time, might be better to just do it in the `__init__`

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -204,14 +254,82 @@ def __init__(
...
+            )
+        )
+
+        metadata, ephemeral_cert = await asyncio.gather(metadata_task, ephemeral_task)
+
+        PEM = OpenSSL.crypto.FILETYPE_PEM
```
kurtisvg: Does this need a variable? Or can it be inline?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -204,14 +254,82 @@ def __init__(
...
+
+        pkey = OpenSSL.crypto.load_privatekey(PEM, self._priv_key)
```
kurtisvg: `pkey` -> `public_key`

A comment for this section explain what is doing on wouldn't go amiss. (In fact, it might make it easier to read this function if you move this part into it's own function)

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -204,14 +254,82 @@ def __init__(
...
+
+        ctx = OpenSSL.SSL.Context(OpenSSL.SSL.TLSv1_2_METHOD)
+        ctx.use_privatekey(pkey)
+        ctx.use_certificate(public_cert)
+        ctx.check_privatekey()
+        ctx.get_cert_store().add_cert(trusted_cert)
+
+        instance_data = {"ssl_context": ctx, "ip_addresses": metadata["ip_addresses"]}
+
+        return instance_data
```
kurtisvg: It doesn't look like this task is self scheduling any longer.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -230,3 +348,72 @@ def _auth_init(self):
...
+
+        :rtype: asyncio.Task
+        :returns: A Task representing _get_instance_data.
+        """
+        print("scheduling")
+
+        try:
+            await asyncio.sleep(delay)
+        except asyncio.CancelledException:
+            return None
```
kurtisvg: A log statement here might be useful

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -204,14 +254,82 @@ def __init__(
 
         data = {"public_key": pub_key}
 
-        resp = await client_session.post(
-            url, headers=headers, json=data, raise_for_status=True
-        )
+        if client_session is not None:
```
kurtisvg: Should client_session ever be none?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -230,3 +348,72 @@ def _auth_init(self):
...
+        fut.set_result(object)
+        return fut
+
+    def connect(self) -> OpenSSL.SSL.Connection:
+        """A method that returns an OpenSSL connection to the database.
+
+        :rtype: OpenSSl.SSL.Connection
+        :returns: An OpenSSL connection to the primary IP of the database.
+        """
+        instance_data = self._current_instance_data.result()
```
kurtisvg: Should result have a timeout ?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -230,3 +348,72 @@ def _auth_init(self):
...
+        ctx = instance_data["ssl_context"]
+        ip_addr = instance_data["ip_addresses"]["PRIMARY"]
```
kurtisvg: At some point, the user should send the ip type they want to use as a parameter

tests/test_instanceconnectionmanager_unit.py
```diff
@@ -14,14 +14,13 @@
 limitations under the License.
 """
 
-import pytest
-import aiohttp
+import pytest  # noqa F401
```
kurtisvg: If you add an exception, add a reason why

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: Opened an issue to address this: #16

fatcat2: For now I'll just have a keyword argument that defaults to primary.

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: Created an issue in #17

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -207,11 +260,104 @@ def __init__(
...
+        return ctx
+
+    def _update_current(self, future) -> None:
+        """A threadsafe way to update the current instance data and the
+        future instance data. Only meant to be called as a callback.
+
+        :type future: asyncio.Future
+        :param future: The future passed in by add_done_callback.
+        """
+        logging.debug("Entered _threadsafe_refresh")
```
kurtisvg: update this logging statement

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -54,14 +64,29 @@ class InstanceConnectionManager:
...
+
     _priv_key: str = None
     _pub_key: str = None
 
+    _current: concurrent.futures.Future = None
+    _next: concurrent.futures.Future = None
+
+    _delay: int = 15
+
+    _logger: logging.Logger = None
```
kurtisvg: This can be a module level variable. 

Each manager_instance is using the same name, so it's actually returning the same object. It is deceptive to assign it when the manager_instance object is created. 

Just put this (after imports and before class):
```python
logger = logging.getLogger(__name__)
```

then log like this:

```python
logger.debug("So python much idiomatic")
```

Even inside the class - the variable is module wide so all of the manager_instances will use the same one.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -78,12 +103,44 @@ def __init__(
...
+        self._mutex = threading.Lock()
+
+        async def create_client_session():
+            return aiohttp.ClientSession()
+
+        self._client_session = asyncio.run_coroutine_threadsafe(
+            create_client_session(), loop=self._loop
+        ).result()
+
+        logging.debug("Updating instance data")
```
kurtisvg: Make sure to change these 
```suggestion
        logger.debug("Updating instance data")
```

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -207,11 +268,104 @@ def __init__(
...
+
+        ctx = OpenSSL.SSL.Context(OpenSSL.SSL.TLSv1_2_METHOD)
+        ctx.use_privatekey(private_key)
+        ctx.use_certificate(public_cert)
+        ctx.check_privatekey()
+        ctx.get_cert_store().add_cert(trusted_cert)
+
+        return ctx
+
+    def _update_current(self, future) -> None:
```
kurtisvg: Should this have a typing on future?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -207,11 +268,104 @@ def __init__(
...
+        """A threadsafe way to update the current instance data and the
+        future instance data. Only meant to be called as a callback.
+
+        :type future: asyncio.Future
+        :param future: The future passed in by add_done_callback.
+        """
+        logging.debug("Entered _update_current")
+        with self._mutex:
+            self._current = future.result()
```
kurtisvg: Do we want the result of future? or the value of future?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -54,14 +65,26 @@ class InstanceConnectionManager:
...
+    _credentials: Credentials = None
+
     _instance_connection_string: str = None
+    _instance: str = None
     _project: str = None
     _region: str = None
-    _instance: str = None
-    _credentials: Credentials = None
+
+    _metadata: Dict[str, Union[Dict, str]] = None
```
kurtisvg: Remove _metadata

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -54,14 +65,26 @@ class InstanceConnectionManager:
...
+
     _priv_key: str = None
     _pub_key: str = None
 
+    _current: concurrent.futures.Future = None
```
kurtisvg: Pair lock with `_current` and `_next`

kurtisvg: Future[type]

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -54,14 +74,31 @@ class InstanceConnectionManager:
...
-    _credentials: Credentials = None
+
     _priv_key: str = None
     _pub_key: str = None
 
+    _lock: threading.Lock = None
+    _current: concurrent.futures.Future = None
+    _next: concurrent.futures.Future = None
+
+    _delay: int = 15
```
kurtisvg: This field is only used in one place - can it be inline? Do you really want it to be 15?

noxfile.py
```diff
@@ -39,7 +39,7 @@ def lint(session):
     session.run("flake8", "google", "tests")
 
 
-@nox.session(python="3.6")
+@nox.session(python="3.7")
```
kurtisvg: Do we want to change this?

noxfile.py
```diff
@@ -66,6 +66,7 @@ def default(session):
         "--cov-config=.coveragerc",
         "--cov-report=",
         "--cov-fail-under=0",
+        # "--concmode=mthread",
```
kurtisvg: Does this line still need to be there?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -230,3 +389,72 @@ def _auth_init(self):
...
+        fut.set_result(object)
+        return fut
+
+    def connect(self,) -> OpenSSL.SSL.Connection:
+        """A method that returns an OpenSSL connection to the database.
+
+        :rtype: OpenSSl.SSL.Connection
+        :returns: An OpenSSL connection to the primary IP of the database.
+        """
+        instance_data: InstanceMetadata = self._current_instance_data.result()
```
kurtisvg: If we are going to use a lock on `self._current` and `self._next` we need to make sure to acquire the lock every time we read/write to these attributes.

You should do something like this: 
```python
with self._lock:
    instance_data = self._current_instance_data
```

It also looks like there are some attributes being mixed up here - is it `self._current` or `self._current_instance_data`?

tests/test_instanceconnectionmanager_unit.py
```diff
@@ -102,22 +122,47 @@ def test_InstanceConnectionManager_get_metadata():
...
     )
+
+
+def test_InstanceConnectionManager_perform_refresh():
+    """
+    Test to check whether _get_perform works as described given valid
+    conditions.
+    """
+    try:
+        connect_string = os.environ["INSTANCE_CONNECTION_NAME"]
```
kurtisvg: Why is this function using a local copy of this variable?

noxfile.py
fatcat2: Probably not if we switch to Kokoro in the future

PR #14, opened 2019-07-18
Converting _get_metadata and _get_ephemeral to asynchronous functions

fatcat2:
The purpose of this PR is to convert _get_metadata and _get_ephemeral to asynchronous functions.

FYI: @kurtisvg, @jkwlui

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 160
--- a/.flake8
+++ b/.flake8
@@ -1,2 +1,13 @@
 [flake8]
-max-line-length = 160
+ignore = E203, E266, E501, W503
+exclude =
+  # Exclude generated code.
+  **/proto/**
+  **/gapic/**
+  *_pb2.py
+
+  # Standard linting exemptions.
+  __pycache__,
+  .git,
+  *.pyc,
+  conf.py
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -78,7 +78,7 @@ class InstanceConnectionManager:
                 "Arg instance_connection_string must be in "
                 + "format: project:region:instance."
             )
-
+        self._loop = loop
         self._auth_init()
         self._priv_key, self._pub_key = generate_keys()
 
@@ -149,7 +149,6 @@ class InstanceConnectionManager:
             TypeError: If one of the arguments passed in is None.
         """
 
-        print(type(project), type(instance), type(pub_key))
         if (
             not isinstance(credentials, Credentials)
             or not isinstance(project, str)
@@ -169,9 +168,8 @@ class InstanceConnectionManager:
             "Content-Type": "application/json",
         }
 
-        # TODO(ryachen@) Add checks to ensure service object is valid.
         url = (
-            "https://www.googleapis.com/sql/v1beta4/projects/%s/instances/%s/createEphemeral"
+            "https://www.googleapis.com/sql/v1beta4/projects/%s/instances/%s/createEphemeral"  # noqa url too long
             % (project, instance)
         )
 
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -171,7 +171,7 @@ class InstanceConnectionManager:
         }
 
         url = (
-            "https://www.googleapis.com/sql/v1beta4/projects/%s/instances/%s/createEphemeral"  # noqa url too long
+            "https://www.googleapis.com/sql/v1beta4/projects/%s/instances/%s/createEphemeral"
             % (project, instance)
         )
 
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -95,7 +95,7 @@ class InstanceConnectionManager:
 
         :type credentials: google.oauth2.service_account.Credentials
         :param service:
-            A service object created from the google-auth Python library.
+            A credentials object created from the google-auth Python library.
             Must have the SQL Admin API scopes. For more info check out
             https://google-auth.readthedocs.io/en/latest/.
 
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -19,7 +19,7 @@ import aiohttp
 import googleapiclient
 import googleapiclient.discovery
 import google.auth
-from google.oauth2.service_account import Credentials
+from google.auth.credentials import Credentials
 from google.cloud.sql.connector.utils import generate_keys
 import google.auth.transport.requests
 import json
@@ -87,7 +87,10 @@ class InstanceConnectionManager:
 
     @staticmethod
     async def _get_metadata(
-        credentials: Credentials, project: str, instance: str
+        client_session: aiohttp.ClientSession,
+        credentials: Credentials,
+        project: str,
+        instance: str,
     ) -> Dict[str, Union[Dict, str]]:
         """Requests metadata from the Cloud SQL Instance
         and returns a dictionary containing the IP addresses and certificate
@@ -140,11 +143,8 @@ class InstanceConnectionManager:
             project, instance
         )
 
-        ret_json = None
-
-        async with aiohttp.ClientSession() as session:
-            async with session.get(url, headers=headers) as resp:
-                ret_json = await resp.text()
+        resp = await client_session.get(url, headers=headers)
+        ret_json = await resp.text()
 
         ret_dict = json.loads(ret_json)
 
@@ -158,7 +158,11 @@ class InstanceConnectionManager:
 
     @staticmethod
     async def _get_ephemeral(
-        credentials: Credentials, project: str, instance: str, pub_key: str
+        client_session: aiohttp.ClientSession,
+        credentials: Credentials,
+        project: str,
+        instance: str,
+        pub_key: str,
     ) -> str:
         """Asynchronously requests an ephemeral certificate from the Cloud SQL Instance.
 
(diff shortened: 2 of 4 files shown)


google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -122,27 +124,48 @@ def _get_metadata(
...
+        }
+
+        url = "https://www.googleapis.com/sql/v1beta4/projects/%s/instances/%s" % (
+            project,
+            instance,
+        )
+
+        ret_json = None
+
+        async with aiohttp.ClientSession() as session:
```
kurtisvg: We probably want to reuse sessions to take advantage of connection pooling.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -122,27 +124,48 @@ def _get_metadata(
...
+            credentials.refresh(request)
+
+        bearer_token = credentials.token
+
+        headers = {
+            "Authorization": "Bearer %s" % bearer_token,
+            "Content-Type": "application/json",
+        }
+
+        url = "https://www.googleapis.com/sql/v1beta4/projects/%s/instances/%s" % (
```
kurtisvg: Can we use `format` instead `%`?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -122,27 +124,48 @@ def _get_metadata(
...
+        ret_dict = json.loads(ret_json)
 
-        # Map IP addresses to type.
-        ip_map = {ip["type"]: ip["ipAddress"] for ip in res["ipAddresses"]}
+        server_ca_cert = ret_dict["serverCaCert"]["cert"]
 
-        metadata = {"ip_addresses": ip_map, "server_ca_cert": serverCaCert}
+        ip_map = {ip["type"]: ip["ipAddress"] for ip in ret_dict["ipAddresses"]}
+
+        metadata = {"ip_addresses": ip_map, "server_ca_cert": server_ca_cert}
```
kurtisvg: can `ip_map` and `server_ca_cert` be moved inline?

requirements.txt
```diff
@@ -1,4 +1,6 @@
+aiohttp==3.5.4
 cryptography==2.7
 PyMySQL==0.9.3
 pytest==4.6.3
+Requests==2.22.0
```
kurtisvg: Why do we need requests?

tests/test_instanceconnectionmanager_unit.py
```diff
@@ -46,3 +47,68 @@ def test_InstanceConnectionManager_wrong_connection_string():
     loop = asyncio.new_event_loop()
     with pytest.raises(CloudSQLConnectionError):
         InstanceConnectionManager("test-project:test-region", loop)
+
+
+def test_InstanceConnectionManager_get_ephemeral():
```
kurtisvg: Ideally this test would mock the response somehow, and we would have some system tests somewhere that would test things end to end.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -76,24 +78,24 @@ def __init__(
...
         """Requests metadata from the Cloud SQL Instance
         and returns a dictionary containing the IP addresses and certificate
         authority of the Cloud SQL Instance.
 
-        :type service: googleapiclient.discovery.Resource
+        :type credentials: google.oauth2.service_account.Credentials
         :param service:
-            A service object created from the Google Python API client library.
-            Must be using the SQL Admin API. For more info check out
+            A service object created from the google-auth Python library.
```
busunkim96:
```suggestion
            A credentials object created from the google-auth Python library.
```

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -76,24 +78,24 @@ def __init__(
...
+            Must be have the SQL Admin API scopes. For more info check out
             https://github.com/googleapis/google-api-python-client.
```
busunkim96: https://github.com/googleapis/google-auth-library-python or https://google-auth.readthedocs.io/en/latest/

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -15,12 +15,14 @@
 """
 
 import asyncio
+import aiohttp
 import googleapiclient
 import googleapiclient.discovery
-from googleapiclient.discovery import Resource
 import google.auth
-from google.auth.credentials import Credentials
+from google.oauth2.service_account import Credentials
```
busunkim96: Are you required to use service account credentials?

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 160
```
crwilcox: https://github.com/googleapis/google-cloud-python/blob/master/container/.flake8 should be used as a template.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -76,24 +78,24 @@ def __init__(
...
         and returns a dictionary containing the IP addresses and certificate
         authority of the Cloud SQL Instance.
 
-        :type service: googleapiclient.discovery.Resource
+        :type credentials: google.oauth2.service_account.Credentials
         :param service:
-            A service object created from the Google Python API client library.
-            Must be using the SQL Admin API. For more info check out
+            A service object created from the google-auth Python library.
+            Must be have the SQL Admin API scopes. For more info check out
```
crwilcox: 'must be have' is an odd phrasing.

google/cloud/sql/connector/InstanceConnectionManager.py
crwilcox: Based on the launch of this, if we only do 3.6+ this could start using f-strings: https://www.python.org/dev/peps/pep-0498/

Also, setup.py likely should specify which versions this supports. My guess though is the plan is 3.5+ since 3.5 won't EOL for a bit yet.

requirements.txt
fatcat2: It's required for `google.auth.transport.requests`, which is used for refreshing credentials if they're not valid.

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: No, just strongly encouraged. I'll change it back to `google.auth.credentials`.

tests/test_instanceconnectionmanager_unit.py
fatcat2: I'll make the changes to these tests in another PR, but it will be done

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -157,21 +186,36 @@ def _get_ephemeral(service, project, instance, pub_key):
...
+        if not credentials.valid:
+            request = google.auth.transport.requests.Request()
+            credentials.refresh(request)
 
-        request = service.sslCerts().createEphemeral(
-            project=project, instance=instance, body={"public_key": pub_key}
+        bearer_token = credentials.token
+
+        headers = {
+            "Authorization": "Bearer %s" % bearer_token,
```
kurtisvg: I think `format()` is preferred over `%`.

I think bearer_token can be moved inline.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -157,21 +186,36 @@ def _get_ephemeral(service, project, instance, pub_key):
...
+
+        url = "https://www.googleapis.com/sql/v1beta4/projects/{}/instances/{}/createEphemeral".format(
+            project, instance
         )
-        response = request.execute()
 
-        return response["cert"]
+        data = {"public_key": pub_key}
+
+        resp = await client_session.post(url, headers=headers, json=data)
```
kurtisvg: What happens if this fails? do you need to call `raise_for_status`?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -157,21 +186,36 @@ def _get_ephemeral(service, project, instance, pub_key):
...
+        ret_json = await resp.text()
```
kurtisvg:
```suggestion
        resp_data = json.loads(await resp.text())
```

requirements.txt
kurtisvg: Just to clarify - why is requests not a dependency of `google.auth.transport.requests` then?

tests/test_instanceconnectionmanager_unit.py
```diff
@@ -46,3 +47,98 @@ def test_InstanceConnectionManager_wrong_connection_string():
...
+    icm = InstanceConnectionManager(connect_string, loop)
+
+    async def create_client_session():
+        return aiohttp.ClientSession()
+
+    fut_client_session = asyncio.ensure_future(create_client_session(), loop=loop)
+
+    icm._loop.run_until_complete(fut_client_session)
+
+    client_session = fut_client_session.result()
```
kurtisvg: L69-79 - Why cant we use `aiohttp.ClientSession(loop=loop)`?

tests/test_instanceconnectionmanager_unit.py
```diff
@@ -46,3 +47,98 @@ def test_InstanceConnectionManager_wrong_connection_string():
...
+        ),
+        loop=loop,
+    )
+
+    icm._loop.run_until_complete(fut)
+    icm._loop.close()
+
+    result = fut.result().split("\n")
+
+    client_session.close()
```
kurtisvg: Since we are using `client_session` in the scope of this function, this should use a `with` statement to ensure it is closed even if any error occurs

tests/test_instanceconnectionmanager_unit.py
fatcat2: According to the `aiohttp` documentation, the `loop` keyword param looks like it's been deprecated

Issue #16, opened 2019-07-26
Finalize design for configuration arguments passed into connect()


fatcat2:
Currently defaulting to use primary, but in the future it would be better to have users specify which IP address they want to connect to.

PR #19, opened 2019-07-31
Added connect method

fatcat2:
Re-working the connect method to return a DB-API connection object rather than an OpenSSL connection.

### DIFF
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -27,11 +27,25 @@ import google.auth.transport.requests
 import json
 from tempfile import NamedTemporaryFile
 import threading
-from typing import Any, Dict, Union
+from typing import Any, Dict, Optional, Union
 
 import logging
 
 logger = logging.getLogger(name=__name__)
+
+# This thread is used to background processing
+_thread: Optional[threading.Thread] = None
+_loop: Optional[asyncio.AbstractEventLoop] = None
+
+
+def _get_loop() -> threading.Thread:
+    if _loop is None:
+        loop = asyncio.new_event_loop()
+        thr = threading.Thread(target=loop.run_forever)
+        thr.start()
+        _thread = thr
+        _loop = loop
+        return _loop
 
 
 class InstanceMetadata:
@@ -56,12 +70,6 @@ class InstanceMetadata:
         self.ca_fileobject.write(server_ca_cert.encode())
         self.cert_fileobject.write(ephemeral_cert.encode())
         self.key_fileobject.write(private_key)
-
-        self.filepaths = {
-            "ca": self.ca_fileobject.name,
-            "cert": self.cert_fileobject.name,
-            "key": self.key_fileobject.name,
-        }
 
 
 class CloudSQLConnectionError(Exception):
@@ -426,6 +434,8 @@ class InstanceConnectionManager:
         """
         logger.debug("Entered connect method")
 
+        # Host and ssl options come from the certificates and metadata, so we don't
+        # want the user to specify them.
         kwargs.pop("host", None)
         kwargs.pop("ssl", None)
 
@@ -453,8 +463,8 @@ class InstanceConnectionManager:
         ca_filepath: str,
         cert_filepath: str,
         key_filepath: str,
-        ip_addresses: Dict[str, str],
-        username: str,
+        ip_address: str,
+        user: str,
         **kwargs
     ):
         """Helper function to create a pymysql DB-API connection object.
@@ -480,7 +490,9 @@ class InstanceConnectionManager:
         try:
             import pymysql
         except ImportError:
-            raise ImportError("Driver pymysql not installed.")
+            raise ImportError(
+                'Unable to import module "pymysql." Please install and try again.'
+            )
 
         ssl_dict = {
             "ssl": {"ca": ca_filepath, "cert": cert_filepath, "key": key_filepath}
@@ -492,14 +504,14 @@ class InstanceConnectionManager:
             )
         )
 
-        return pymysql.connect(host=ip_addresses["PRIMARY"], ssl=ssl_dict, **kwargs)
+        return pymysql.connect(host=ip_address, ssl=ssl_dict, user=user, **kwargs)
 
     def _connect_with_pg8000(
         self,
         ca_filepath: str,
         cert_filepath: str,
         key_filepath: str,
-        ip_addresses: Dict[str, str],
+        ip_address: str,
         username: str,
         **kwargs
     ):
@@ -526,7 +538,9 @@ class InstanceConnectionManager:
         try:
             import pg8000
         except ImportError:
-            raise ImportError("Driver pg8000 not installed.")
+            raise ImportError(
+                'Unable to import module "pg8000." Please install and try again.'
+            )
 
         ssl_dict = {
             "ca_certs": ca_filepath,
@@ -540,4 +554,4 @@ class InstanceConnectionManager:
             )
         )
 
-        return pg8000.connect(username, ip_addresses["PRIMARY"], ssl=ssl_dict, **kwargs)
+        return pg8000.connect(username, ip_address, ssl=ssl_dict, **kwargs)
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -40,6 +40,12 @@ class InstanceMetadata:
     private_key: str
     server_ca_cert: str
 
+    ca_fileobject: NamedTemporaryFile
+    cert_fileobject: NamedTemporaryFile
+    key_fileobject: NamedTemporaryFile
+
+    filepaths: Dict[str, str] = {}
+
     def __init__(
         self,
         ephemeral_cert: str,
@@ -51,6 +57,20 @@ class InstanceMetadata:
         self.server_ca_cert = server_ca_cert
         self.ephemeral_cert = ephemeral_cert
         self.private_key = private_key
+
+        self.ca_fileobject = NamedTemporaryFile(suffix=".pem")
+        self.cert_fileobject = NamedTemporaryFile(suffix=".pem")
+        self.key_fileobject = NamedTemporaryFile(suffix=".pem")
+
+        self.ca_fileobject.write(server_ca_cert.encode())
+        self.cert_fileobject.write(ephemeral_cert.encode())
+        self.key_fileobject.write(private_key)
+
+        self.filepaths = {
+            "ca": self.ca_fileobject.name,
+            "cert": self.cert_fileobject.name,
+            "key": self.key_fileobject.name,
+        }
 
 
 class CloudSQLConnectionError(Exception):
@@ -232,9 +252,8 @@ class InstanceConnectionManager:
     ) -> str:
         """Asynchronously requests an ephemeral certificate from the Cloud SQL Instance.
 
-        
         :type credentials: google.oauth2.service_account.Credentials
-        :param credentials: A credentials object 
+        :param credentials: A credentials object
             created from the google-auth library. Must be
             using the SQL Admin API scopes. For more info, check out
             https://google-auth.readthedocs.io/en/latest/.
@@ -244,8 +263,8 @@ class InstanceConnectionManager:
 
         :type instance: str
         :param instance: A string representing the name of the instance.
-        
-        :type pub_key: 
+
+        :type pub_key:
         :param str: A string representing PEM-encoded RSA public key.
 
         :rtype: str
@@ -335,7 +354,8 @@ class InstanceConnectionManager:
         logger.debug("Entered _update_current")
         with self._lock:
             self._current = future
-            self._next = self._loop.create_task(self._schedule_refresh(5))
+            # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
+            self._next = self._loop.create_task(self._schedule_refresh(55 * 60))
 
     def _auth_init(self) -> None:
         """Creates and assigns a Google Python API service object for
@@ -424,14 +444,6 @@ class InstanceConnectionManager:
         with self._lock:
             instance_data: InstanceMetadata = self._current.result()
 
-        ssl_ca = NamedTemporaryFile(suffix=".pem")
-        ssl_cert = NamedTemporaryFile(suffix=".pem")
-        ssl_key = NamedTemporaryFile(suffix=".pem")
-
-        ssl_ca.write(instance_data.server_ca_cert.encode())
-        ssl_cert.write(instance_data.ephemeral_cert.encode())
-        ssl_key.write(instance_data.private_key)
-
         try:
             connector = {
                 "pymysql": self._connect_with_pymysql,
@@ -441,15 +453,16 @@ class InstanceConnectionManager:
             raise KeyError("Driver {} is not supported.".format(driver))
 
         return connector(
-            ssl_ca.name,
-            ssl_cert.name,
-            ssl_key.name,
+            instance_data.filepaths["ca"],
+            instance_data.filepaths["cert"],
+            instance_data.filepaths["key"],
             instance_data.ip_addresses,
             username,
             **kwargs
         )
 
     def _connect_with_pymysql(
+        self,
         ca_filepath: str,
         cert_filepath: str,
         key_filepath: str,
@@ -498,6 +511,7 @@ class InstanceConnectionManager:
         )
 
     def _connect_with_pg8000(
+        self,
         ca_filepath: str,
         cert_filepath: str,
         key_filepath: str,
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -21,34 +21,19 @@ from google.cloud.sql.connector.utils import generate_keys, write_to_file
 import asyncio
 import aiohttp
 import concurrent
-from cryptography.x509 import load_pem_x509_certificate
-from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives.serialization import Encoding
 import google.auth
 from google.auth.credentials import Credentials
 import google.auth.transport.requests
 import json
+import ssl
+import socket
 from tempfile import NamedTemporaryFile
 import threading
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Union
 
 import logging
 
 logger = logging.getLogger(name=__name__)
-
-# This thread is used to background processing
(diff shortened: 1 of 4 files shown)


google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -29,21 +29,27 @@
...
 from typing import Any, Dict, Union
 
 import logging
 
 logger = logging.getLogger(name=__name__)
 
+  
 
 class InstanceMetadata:
+    ephemeral_cert: str
```
kurtisvg: These should probably have filepath in the name, so it is clear they are files? 

You probably also want to store a reference to the FileObject itself in this class, so that it doesn't get garbage collected before it's supposed to.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +403,56 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
         fut.set_result(object)
         return fut
 
-    def connect(self,) -> OpenSSL.SSL.Connection:
+    def connect(self, driver: str, username:str=None, **kwargs) -> OpenSSL.SSL.Connection:
```
kurtisvg: Do you need to update this docstring?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +403,56 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+        
+        ssl_ca = NamedTemporaryFile(suffix=".pem")
+        ssl_cert = NamedTemporaryFile(suffix=".pem")
+        ssl_key = NamedTemporaryFile(suffix=".pem")
+
+        ssl_ca.write(instance_data.server_ca_cert.encode())
+        ssl_cert.write(instance_data.ephemeral_cert.encode())
+        ssl_key.write(instance_data.private_key)
+
+        if driver is "pg8000":
```
kurtisvg: Don't use `is` to compare objects (is is more like `===`)
```suggestion
        if driver == "pg8000":
```

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +403,56 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+            try:
+                import pg8000
+                logger.debug("Successfully imported %s", driver)
```
kurtisvg: Is this debug necessary? The alternative is an exception is thrown, so the very next statement indicates that it was thrown successfully.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +403,56 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+                ssl_dict = {
+                    "ca_certs": ssl_ca.name,
+                    "certfile": ssl_cert.name,
+                    "keyfile": ssl_key.name
+                }
+                logger.debug("Temporary files: {}, {} and {}".format(ssl_ca.name, ssl_cert.name, ssl_key.name))
+                
+                return pg8000.connect(username, instance_data.ip_addresses["PRIMARY"], ssl=ssl_dict, **kwargs)
+            except ImportError as e:
+                raise e
```
kurtisvg: You shouldn't catch an exception if you are just going to throw it again unchanged (it's an antipattern). You should show a message explaining to the use that they need to have the driver installed if they want to use it. 

Additionally, the scope of this try/except is too big - as a general rule, you should limit to code that throws the error to avoid catching something else.

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: The `InstanceMetadata` object is created in `_get_instance_data`, I can just move the creation of the temporary files into `_get_instance_data` or even into the constructor of the `InstanceMetadata` object.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +425,136 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+        }
+
+        logger.debug(
+            "Temporary files: {}, {} and {}".format(
+                ca_filepath, cert_filepath, key_filepath
+            )
+        )
+
+        return pymysql.connect(
+            host=ip_addresses["PRIMARY"], user=username, ssl=ssl_dict, **kwargs
```
kurtisvg: What happens if `host` is passed in kwargs? Does it raise an error?

google/cloud/sql/connector/connector.py
```diff
@@ -44,4 +49,8 @@ def connect(instance_connection_string, **kwargs):
     #
     # Return a DBAPI connection
 
-    raise NotImplementedError()
+    loop = asyncio.new_event_loop()
+    thr = Thread(target=loop.run_forever)
```
kurtisvg: This `loop`/`thread` combo would be a good target for a fixture.

```python
@fixture
def async_loop():
     loop = asyncio.new_event_loop()
     thread = Thread(target=loop.run_forever)
     thread.start()
     yield loop
     loop.stop()
     thread.join()
```

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: It should, since we want to override host no matter what

google/cloud/sql/connector/connector.py
fatcat2: The fixture would be in the tests, right/

fatcat2:
This PR handles #16

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -38,12 +35,42 @@
...
+
+        self.ca_fileobject = NamedTemporaryFile(suffix=".pem")
+        self.cert_fileobject = NamedTemporaryFile(suffix=".pem")
+        self.key_fileobject = NamedTemporaryFile(suffix=".pem")
+
+        self.ca_fileobject.write(server_ca_cert.encode())
+        self.cert_fileobject.write(ephemeral_cert.encode())
+        self.key_fileobject.write(private_key)
+
+        self.filepaths = {
```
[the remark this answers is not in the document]
kurtisvg: If we aren't using these properties in the object, can we remove them? I think the "fileobjects" are the only ones being used, so can we remove the rest?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +416,128 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
-        :rtype: OpenSSl.SSL.Connection
-        :returns: An OpenSSL connection to the primary IP of the database.
+        :type driver: str
+        :param driver: A string representing the driver. e.g. "pg8000" or "pymysql"
+
+        :returns: A DB-API connection to the primary IP of the database.
         """
+        logger.debug("Entered connect method")
+
+        kwargs.pop("host", None)
```
kurtisvg: nit: A comment here clarifying why you are popping these would be good

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +416,128 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+        :param ip_addresses: A Dictionary containing the different IP addresses
+            of the Cloud SQL instance.
+
+        :rtype: pymysql.Connection
+        :returns: A PyMySQL Connection object for the Cloud SQL instance.
+        """
+        try:
+            import pymysql
+        except ImportError:
+            raise ImportError("Driver pymysql not installed.")
```
kurtisvg:
```suggestion
            raise ImportError('Unable to import module "pymysql". Please install and try again.')
```

google/cloud/sql/connector/connector.py
kurtisvg: Yes - sorry that wasn't clear.

google/cloud/sql/connector/connector.py
```diff
@@ -44,4 +49,8 @@ def connect(instance_connection_string, **kwargs):
...
+    thr.start()
```
kurtisvg: 2 problems with this: 
1. A reference to Thread isn't being held, so it could be garbage collected
2. We don't want a new Thread/loop for each connect, we want to reuse the same thread/loop for efficiency

You can solve this by using a module level variable:
```python
# This thread is used to background processing
_thread: Optional[Thread] = None
_loop: Optional[asyncio.Loop] = None
```

And a function to memoize:
```python
def _get_loop() -> Thread:
   if _loop is None:
     # init thread and loop
   return _loop
```

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -38,12 +37,32 @@
...
+        private_key: str,
+        server_ca_cert: str,
+    ):
+        self.ip_address = ip_address
+
+        self.ca_fileobject = NamedTemporaryFile(suffix=".pem")
+        self.cert_fileobject = NamedTemporaryFile(suffix=".pem")
+        self.key_fileobject = NamedTemporaryFile(suffix=".pem")
+
+        write_to_file(server_ca_cert, ephemeral_cert, private_key)
```
kurtisvg: You need to remove this

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -38,12 +37,32 @@
...
+
+        self.ca_fileobject.write(server_ca_cert.encode())
+        self.ca_fileobject.seek(0)
```
kurtisvg: Why do you call seek? This section would be a good place to put a comment explaining what/why you are doing this (since it's not obvious from reading the code)

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -201,7 +220,9 @@ def __del__(self):
             project, instance
         )
 
-        logging.getLogger(__name__).debug("Requesting metadata")
+        logger.debug(url)
```
kurtisvg: This debug looks unnecessary, might be best to remove it

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +419,145 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
-        ssl_connection = OpenSSL.SSL.Connection(ctx, s)
-        ssl_connection.connect((ip_addr, 3307))
-        ssl_connection.do_handshake()
-
-        return ssl_connection
+
+        try:
+            connector = {
+                "pymysql": self._connect_with_pymysql,
+                "pg8000": self._connect_with_pg8000,
```
kurtisvg: Does pg8000 work? If not, maybe best to leave out for now

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +419,145 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+        """
+        try:
+            import pymysql
+        except ImportError:
+            raise ImportError(
+                'Unable to import module "pymysql." Please install and try again.'
+            )
+
+        # ssl_dict = {
+        # "ssl": {"ca": ca_filepath, "cert": cert_filepath, "key": key_filepath, 'check_hostname': False}
```
kurtisvg: Does this comment need to be here? If so, consider formatting it properly. If not, remove it.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +419,145 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+        # }
+
+        logger.debug(
```
kurtisvg: I would consider not writing out the location of these files in the log - might be a security issue?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +419,145 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+            "Temporary files: {}, {} and {}".format(
+                ca_filepath, cert_filepath, key_filepath
+            )
+        )
+
+        ctx = ssl.SSLContext()
```
kurtisvg: Can we reuse the SSLContext for each connection? If so, maybe this should go back in InstanceMetadata

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +419,145 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
+        ctx.load_cert_chain(cert_filepath, keyfile=key_filepath)
+        ctx.load_verify_locations(cafile=ca_filepath)
+
+        sock = ctx.wrap_socket(
+            socket.create_connection((ip_address, 3307)), server_hostname=ip_address
+        )
+
+        conn = pymysql.Connection(host=ip_address, defer_connect=True, **kwargs)
+
+        conn.connect(sock)
```
kurtisvg: This section could use some more comments about why this is necessary

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -38,12 +37,41 @@
...
+
+        self.ca_fileobject.seek(0)
+        self.cert_fileobject.seek(0)
+        self.key_fileobject.seek(0)
+
+        self.context = ssl.SSLContext()
+        self.context.load_cert_chain(self.cert_fileobject.name, keyfile=self.key_fileobject.name)
+        self.context.load_verify_locations(cafile=self.ca_fileobject.name)
+
+        self.ca_fileobject.seek(0)
```
kurtisvg: nit: I think these `seek` are unnecessary

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -38,12 +37,41 @@
...
+        # Write each file and reset to beginning
+        self.ca_fileobject.write(server_ca_cert.encode())
+        self.cert_fileobject.write(ephemeral_cert.encode())
+        self.key_fileobject.write(private_key)
+
+        self.ca_fileobject.seek(0)
+        self.cert_fileobject.seek(0)
+        self.key_fileobject.seek(0)
+
+        self.context = ssl.SSLContext()
```
kurtisvg: If we are creating the context here, can we make the fileobjects private? They shouldn't be used anywhere else, right?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -38,12 +37,43 @@
...
+    def __init__(
+        self,
+        ephemeral_cert: str,
+        ip_address: str,
+        private_key: str,
+        server_ca_cert: str,
+    ):
+        self.ip_address = ip_address
+
+        self.ca_fileobject = NamedTemporaryFile(suffix=".pem")
```
crwilcox: The docs for `NamedTemporaryFiles` warn that they cannot be opened again on Windows. https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

The later lines concern me (`load_verify_locations`) as I am pretty sure that is going to attempt to open the file again.

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -364,6 +357,7 @@ def _update_current(self, future: concurrent.futures.Future) -> None:
         logger.debug("Entered _update_current")
         with self._lock:
             self._current = future
+            # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
             self._next = self._loop.create_task(self._schedule_refresh(55 * 60))
```
crwilcox: It may make sense to pull this into a private constant to avoid the inline magic number

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -379,12 +373,7 @@ def _auth_init(self) -> None:
             ]
         )
 
-        cloudsql = googleapiclient.discovery.build(
-            "sqladmin", "v1beta4", credentials=scoped_credentials
```
crwilcox: Similar to other number, putting this version in a constant might be useful, especially if there is/could be multiple builds ever? If not, it is less of a concern.

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: General windows incompatibility is being tracked in #22

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -21,14 +21,13 @@
 import asyncio
 import aiohttp
 import concurrent
-import googleapiclient
-import googleapiclient.discovery
 import google.auth
 from google.auth.credentials import Credentials
 import google.auth.transport.requests
 import json
-import OpenSSL
```
crwilcox: Are there updates needed to setup.py requirements?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -364,6 +363,7 @@ def _update_current(self, future: concurrent.futures.Future) -> None:
         logger.debug("Entered _update_current")
         with self._lock:
             self._current = future
+            # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
             self._next = self._loop.create_task(self._schedule_refresh(55 * 60))
```
crwilcox: Can this use the constant you created?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -439,19 +434,66 @@ def immediate_future(object: Any) -> concurrent.futures.Future:
...
 
-    def connect(self,) -> OpenSSL.SSL.Connection:
-        """A method that returns an OpenSSL connection to the database.
+    def connect(self, driver: str, **kwargs) -> Any:
+        """A method that returns a DB-API connection to the database.
 
-        :rtype: OpenSSl.SSL.Connection
-        :returns: An OpenSSL connection to the primary IP of the database.
+        :type driver: str
+        :param driver: A string representing the driver. e.g. "pg8000" or "pymysql"
```
crwilcox: It seems the pg8000 support is gone now?

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: Definitely, I'm currently taking care of that in #21

google/cloud/sql/connector/InstanceConnectionManager.py
fatcat2: Yeah, still need to work out the specifics on the postgres drivers

Issue #26, opened 2020-04-17
connector.connect() hangs on exit


shivaal-scio:
Repro:
```bash
python3 -m venv cloudsql_env
source cloudsql_env/bin/activate
pip install git+https://github.com/GoogleCloudPlatform/cloud-sql-python-connector@41e317ecdf6131d933ee455cd68fc4006aac8584

ipython
> from google.cloud.sql.connector import connector
> conn = connector.connect(instance_connection_string=<instance>, driver='pymysql', user=<user>, password=<password>, db=<db>)
> exit
[hangs]
```

This is hanging because the `_loop.run_forever` thread created on [connector.py#L33](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/master/google/cloud/sql/connector/connector.py#L33) is still running in the background. I was able to fix this by setting it as a daemon thread, so that it gets killed when the program exits:

```python
_thread = Thread(target=_loop.run_forever, daemon=True)
```

@crwilcox I believe this was Ryan's intern project, but I'd really appreciate if this could be fixed, even if this is no longer actively maintained. Would you mind making the change? I've confirmed that with them, I can make a connection using this library. (FWIW I believe the creator of the previous issue was not affected because Dataflow might kill the process on its own.)

kurtisvg:
@shivaal-scio You are welcome to open a PR with the change :)

shivaal-scio:
Good point @kurtisvg, I just created https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/27

PR #27, opened 2020-04-17
fix: avoid hang on connector.connect

shivaal-scio:
Fixes https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/issues/26


shivaal-scio:
@googlebot I signed it!

PR #35, opened 2021-01-28
chore: update dependencies to latest versions to get tests passing

shubha-rajan:
Updates dependencies in requirements.txt to latest versions. This will fix test failures due to incompatibility between pytest and pytest-asyncio versions.


shubha-rajan:
The current build failures are because the service account file that Travis is using is no longer working. We can hold off on merging this until Kokoro is set up if that's better.

kurtisvg (approved):

PR #39, opened 2021-02-02
testing: add integration testing for pymysql driver support

shubha-rajan:
Added a system test for using the connector with `pymysql`, similar to the integration tests for the Java connector


kurtisvg (approved):
A couple of the renamed files have some stutter we can probably remove (e.g. "tests/test_something_unit.py" -> "tests/unit/test_something_unit.py"). Can we remove the extra "unit"?

PR #40, opened 2021-02-03
feat: add pg8000 support

shubha-rajan:
Ready for review.
Notes:
* For now, it's installing directly from GitHub. We'll have to wait until the next release of pg8000 to update it. 
* I had to subclass SSLContext to get the Python 3.6 tests to pass. The Python 3.6 fail when adding a [pg8000 specific `request_ssl` attribute](https://github.com/tlocke/pg8000/commit/37dcbe32bc6703dbacebd8f0d8ec94cad946fda7) to a `ssl.SSLContext` object.


google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -42,6 +42,14 @@
 _sql_api_version: str = "v1beta4"
 
 
+class ConnectionSSLContext(ssl.SSLContext):
+    """Subclass of ssl.SS:Context with added request_ssl attribute."""
```
kurtisvg, line 46: Is there a typo here?  I think a little more explanation on why this class is used in the code would be good as well.

kurtisvg (approved):

google/cloud/sql/connector/InstanceConnectionManager.py
shubha-rajan: Good catch on the typo. And yeah, I can expand on this a bit more. Does the class name look ok? I wasn't sure what to call it.

PR #54, opened 2021-03-04
feat: set User-Agent to include version and SQL driver info

shubha-rajan:
Confirmed with @jsimonweb that the new userAgent strings are showing up.


google/cloud/sql/connector/connector.py
```diff
@@ -65,5 +67,6 @@ def connect(instance_connection_string, driver: str, **kwargs):
     # Return a DBAPI connection
 
     loop = _get_loop()
-    icm = InstanceConnectionManager(instance_connection_string, loop)
+    user_agent = f"cloud-sql-python-connector-{driver}/{__version__}"
+    icm = InstanceConnectionManager(instance_connection_string, user_agent, loop)
```
kurtisvg, line 71: Do we need to pass version in as an argument, or can we just import it directly?

google/cloud/sql/connector/connector.py
```diff
@@ -65,5 +67,6 @@ def connect(instance_connection_string, driver: str, **kwargs):
     # Return a DBAPI connection
 
     loop = _get_loop()
-    icm = InstanceConnectionManager(instance_connection_string, loop)
+    user_agent = f"cloud-sql-python-connector-{driver}/{__version__}"
```
kurtisvg, line 70: Is the end goal to have different artifacts for `cloud-sql-python-connector` in each driver? Or will it be just one artifact to download and install? 

If it's one, we should consider the driver as a metadata tag instead?

kurtisvg (changes requested):

google/cloud/sql/connector/connector.py
shubha-rajan: you mean import it directly in the instanceConnectionManager file?

google/cloud/sql/connector/connector.py
shubha-rajan: We could append it as version metadata instead.

Also, I'm actually not sure if we should release one or multiple artifacts. The issue with releasing one artifact is that you end up installing unnecessary drivers when you only need one. Another solution would be to have users install the driver separately but I can see there being a lot of issues with version incompatibilities.

google/cloud/sql/connector/connector.py
shubha-rajan: Refactored this so that we just pass in the driver.

jsimonweb (approved):

kurtisvg (approved):

PR #57, opened 2021-03-08
feat: add pytds support

shubha-rajan:
Add support for [pytds](https://github.com/denisenkom/pytds/) driver

DO NOT MERGE until:

- [x] https://github.com/denisenkom/pytds/pull/120 is merged
- [x] b/179703611 is resolved

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@ aiohttp==3.7.4.post0
 cryptography==3.4.7
 PyMySQL==1.0.2
 pg8000==1.19.2
+python-tds==1.10.0
 pyopenssl==20.0.1
 pytest==6.2.3
 Requests==2.25.1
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -2,6 +2,8 @@ aiohttp==3.7.4.post0
 cryptography==3.4.6
 PyMySQL==1.0.2
 pg8000==1.18.0
+# python-tds==1.10.0
+git+https://github.com/denisenkom/pytds.git@refs/pull/120/merge
```
kurtisvg, line 6: I think I've mentioned this before, but we should probably make these scoped requirements rather than include them all by default.

kurtisvg (approved):

requirements.txt
shubha-rajan: just updated the PR to scope the dependencies in setup.py

PR #59, opened 2021-03-09
fix: generate keys asynchronously

### LINKED ISSUE #15: Generate keys asynchronously
Change _generate_keys to asynchronous

shubha-rajan:
[WIP]
fixes #15

### DIFF
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -30,7 +30,7 @@ import ssl
 import socket
 from tempfile import NamedTemporaryFile
 import threading
-from typing import Any, Dict, Union
+from typing import Any, Dict, Union, Awaitable
 
 import logging
 
@@ -179,7 +179,7 @@ class InstanceConnectionManager:
 
         self._user_agent_string = f"{APPLICATION_NAME}/{version}+{driver_name}"
         self._loop = loop
-        self._key_generation_task = self._loop.create_task(self._get_keys())
+        self._keys: Awaitable = self._loop.create_task(self._get_keys())
         self._auth_init()
         self._lock = threading.Lock()
 
@@ -345,8 +345,7 @@ class InstanceConnectionManager:
         """Asynchronous function to generate and set values for public and
         private keys.
         """
-        self._priv_key, pub_key = await generate_keys(loop=self._loop)
-        self._pub_key = pub_key.decode("UTF-8")
+        self._priv_key, self._pub_key = await generate_keys()
 
     async def _get_instance_data(self) -> InstanceMetadata:
         """Asynchronous function that takes in the futures for the ephemeral certificate
@@ -357,7 +356,7 @@ class InstanceConnectionManager:
             containing the instances IP adresses, a string representing a PEM-encoded private key
             and a string representing a PEM-encoded certificate authority.
         """
-        await self._key_generation_task
+        await self._keys
 
         logger.debug("Creating context")
 
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -30,7 +30,7 @@ import ssl
 import socket
 from tempfile import NamedTemporaryFile
 import threading
-from typing import Any, Dict, Tuple, Union
+from typing import Any, Dict, Union
 
 import logging
 
@@ -179,7 +179,9 @@ class InstanceConnectionManager:
 
         self._user_agent_string = f"{APPLICATION_NAME}/{version}+{driver_name}"
         self._loop = loop
-        self._key_generation_task = asyncio.ensure_future(self._get_keys(), loop=self._loop)
+        self._key_generation_task = asyncio.ensure_future(
+            self._get_keys(), loop=self._loop
+        )
         self._auth_init()
         self._lock = threading.Lock()
 
@@ -340,8 +342,11 @@ class InstanceConnectionManager:
         ret_dict = json.loads(await resp.text())
 
         return ret_dict["cert"]
-    
-    async def _get_keys(self) -> Tuple[bytes, bytes]:
+
+    async def _get_keys(self) -> None:
+        """Asynchronous function to generate and set values for public and
+        private keys.
+        """
         self._priv_key, pub_key = await generate_keys()
         self._pub_key = pub_key.decode("UTF-8")
 
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -351,7 +351,7 @@ class InstanceConnectionManager:
             and a string representing a PEM-encoded certificate authority.
         """
         if not self._priv_key or not self._pub_key:
-            self._priv_key, self._pub_key = self._keys.result()
+            self._priv_key, self._pub_key = await asyncio.wrap_future(self._keys)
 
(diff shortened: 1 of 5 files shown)


google/cloud/sql/connector/utils.py
```diff
@@ -14,13 +14,25 @@
...
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 
 
+def run_function_as_async(func):
+    @wraps(func)
+    async def wrapped_sync_function(*args, **kwargs):
+        partial_func = partial(func, *args, **kwargs)
+        loop = asyncio.get_event_loop()
```
kurtisvg, line 30: one thing we need to be careful here is that I think each ICM has a different asyncio loop being used. Are we sure this is going to be getting the right one?

google/cloud/sql/connector/utils.py
```diff
@@ -14,13 +14,25 @@
...
+        return await loop.run_in_executor(None, partial_func)
```
kurtisvg, line 31: are we using the default executor for the rest of the threads? (if not, maybe we should?)

google/cloud/sql/connector/utils.py
```diff
@@ -56,19 +68,37 @@ def generate_keys():
...
-    public_key = private_key_obj.public_key().public_bytes(
+
+
+@run_function_as_async
+def get_public_key_bytes(private_key_obj):
+    """Helper function to get public key bytes from private key object."""
+    return private_key_obj.public_key().public_bytes(
         encoding=serialization.Encoding.PEM,
         format=serialization.PublicFormat.SubjectPublicKeyInfo,
     )
```
kurtisvg, line 92: I'm a little concern that splitting these up isn't adding a lot of value - they are mostly just formatting, and not sure the overhead of using a background thread will outweigh the cost to format.

kurtisvg (changes requested):

google/cloud/sql/connector/utils.py
shubha-rajan: I think because the function was called with loop.create_task it would get the right loop, but just to be sure, I updated this in the last commit to explicitly pass in a loop in kwargs

google/cloud/sql/connector/utils.py
shubha-rajan: I think this is the only `run_in_executor` call

google/cloud/sql/connector/utils.py
```diff
@@ -14,13 +14,25 @@
...
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 
 
+def run_function_as_async(func):
+    @wraps(func)
+    async def wrapped_sync_function(*args, **kwargs):
+        partial_func = partial(func, *args, **kwargs)
+        loop = kwargs.pop("loop") or asyncio.get_event_loop()
```
kurtisvg, line 30: So after looking back at this code with more context, I think this was better without the explicit loop. It should only be used from an asyncio context, and giving the ability to pass a loop just introduces a chance for mistakes to happen.

kurtisvg (changes requested):

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -351,6 +350,8 @@ async def _get_instance_data(self) -> InstanceMetadata:
             containing the instances IP adresses, a string representing a PEM-encoded private key
             and a string representing a PEM-encoded certificate authority.
         """
+        if not self._priv_key or not self._pub_key:
+            self._priv_key, self._pub_key = await asyncio.wrap_future(self._keys)
```
kurtisvg, line 354: probably don't need to assign these to `self`:
```suggestion
priv_key, pub_key = await self._keys
```

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -179,9 +179,8 @@ def __init__(
 
         self._user_agent_string = f"{APPLICATION_NAME}/{version}+{driver_name}"
         self._loop = loop
+        self._keys: Awaitable = keys
```
kurtisvg, line 182: This should have a specific return type: 
```suggestion
        self._keys: Awaitable[Type1, Type2] = keys
```

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -161,6 +160,7 @@ def __init__(
         self,
         instance_connection_string: str,
         driver_name: str,
+        keys: concurrent.futures.Future,
```
kurtisvg, line 163: `concurrent.futures.Future` isn't Awaitable, so you'll need to make a coro with `loop.call_soon_threadsafe`) and only use it from asyncio-land.

google/cloud/sql/connector/connector.py
```diff
@@ -68,10 +78,11 @@ def connect(instance_connection_string, driver: str, **kwargs):
     # Return a DBAPI connection
 
     loop = _get_loop()
+    keys = _get_keys()
```
kurtisvg, line 81: nit: we only need `_get_keys()` in the 2nd clause of this if statement

kurtisvg (changes requested):

google/cloud/sql/connector/InstanceConnectionManager.py
shubha-rajan: Done

google/cloud/sql/connector/InstanceConnectionManager.py
shubha-rajan: Done

google/cloud/sql/connector/connector.py
shubha-rajan: Done

google/cloud/sql/connector/InstanceConnectionManager.py
shubha-rajan: I'm not sure how to get `call_soon_threadsafe` to work for this. It returns an `asyncio.Handle` which isn't Awaitable.  Any reason why awaiting the wrapped future from the asyncio thread won't work?

kurtisvg: Ah my bad --  I didn't realize this was in the `__init__` signature. So the two options are:

1. Use a future and wrap it as a Task. This makes the future more async compatible but also introduces more overhead (since futures are thread level).

2. Pass in a coro directly and start it with thread_safe. This is less overhead but blurs the lines between thread-land and asyncio-land. 

Not sure if there is a clear winner.

tests/unit/test_utils.py
```diff
@@ -16,22 +16,26 @@
...
+    res1, res2 = await utils.generate_keys()
     assert (res1 is not None) and (res2 is not None)
 
 
-def test_generate_keys_returns_bytes():
+@pytest.mark.asyncio
+async def test_generate_keys_returns_bytes_and_str():
     """
     Test to check if objects produced from the generate_keys() function are of
     type bytes.
```
kurtisvg, line 37:
```suggestion
    the expected types.
```

google/cloud/sql/connector/InstanceConnectionManager.py
kurtisvg: I think this is probably fine.

kurtisvg (approved):

PR #60, opened 2021-03-09
fix: add timeout in InstanceConnectionManager.connect

### LINKED ISSUE #17: Add timeout in InstanceConnectionManager.connect
Currently `InstanceConnectionManager.connect` blocks and waits for the result of `self._current_instance_data` before moving on to connect using the OpenSSL SSLcontext object. Connect needs some way to timeout.

### LINKED ISSUE #61: InstanceConnectionManager.py should be "instance_connection_manager.py"

shubha-rajan:
fixes #17
fixes #61

### DIFF
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -30,6 +30,7 @@ import socket
 from tempfile import NamedTemporaryFile
 from typing import Any, Dict, Union
 
+from functools import partial
 import logging
 
 logger = logging.getLogger(name=__name__)
@@ -488,7 +489,11 @@ class InstanceConnectionManager:
         except KeyError:
             raise KeyError("Driver {} is not supported.".format(driver))
 
-        return connector(instance_data.ip_address, instance_data.context, **kwargs)
+        connect_partial = partial(
+            connector, instance_data.ip_address, instance_data.context, **kwargs
+        )
+
+        return await self._loop.run_in_executor(None, connect_partial)
 
     def _connect_with_pymysql(self, ip_address: str, ctx: ssl.SSLContext, **kwargs):
         """Helper function to create a pymysql DB-API connection object.
--- a/google/cloud/sql/connector/InstanceConnectionManager.py
+++ b/google/cloud/sql/connector/InstanceConnectionManager.py
@@ -156,7 +156,6 @@ class InstanceConnectionManager:
     _priv_key: str = None
     _pub_key: str = None
 
-    _lock: threading.Lock = None
     _current: concurrent.futures.Future = None
     _next: concurrent.futures.Future = None
 
@@ -185,13 +184,13 @@ class InstanceConnectionManager:
         self._loop = loop
         self._keys: Awaitable = asyncio.wrap_future(keys, loop=self._loop)
         self._auth_init()
-        self._lock = threading.Lock()
+        self._priv_key, pub_key = generate_keys()
+        self._pub_key = pub_key.decode("UTF-8")
 
         logger.debug("Updating instance data")
 
-        with self._lock:
-            self._current = self._perform_refresh()
-            self._next = self.immediate_future(self._current)
+        self._current = self._perform_refresh()
+        self._next = self.immediate_future(self._current)
 
     def __del__(self):
         """Deconstructor to make sure ClientSession is closed and tasks have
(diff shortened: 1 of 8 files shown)


google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -43,6 +43,10 @@
 _delay: int = 55 * 60
 _sql_api_version: str = "v1beta4"
 
+# Time in seconds to wait for the result of fetching instance metadata before
+# timing out.
+_instance_metadata_timeout = 30
```
kurtisvg, line 48: IMO, this should be a user configurable timeout that's passed in via the function's signature. 

Is there anything the the DB API spec that defines a timeout arg?

google/cloud/sql/connector/InstanceConnectionManager.py
```diff
@@ -476,7 +480,15 @@ def connect(self, driver: str, **kwargs) -> Any:
...
+                instance_data: InstanceMetadata = self._current.result(
+                    timeout=_instance_metadata_timeout
+                )
+            except concurrent.futures.TimeoutError:
+                raise TimeoutError(
+                    "Failed to fetch instance metadata. "
+                    f"Task timed out after {_instance_metadata_timeout}s"
+                )
 
         connect_func = {
```
kurtisvg, line 493: Seems like the timeout should be passed into the connector function as well?

kurtisvg (changes requested):

google/cloud/sql/connector/InstanceConnectionManager.py
kurtisvg: Interesting thing to think about: do we want the same total timeout for the whole function, or a timeout for each part (I'm guessing the first is what most people expect)

google/cloud/sql/connector/InstanceConnectionManager.py
shubha-rajan: There's nothing in the DB API spec so I think it's driver specific: https://www.python.org/dev/peps/pep-0249/

google/cloud/sql/connector/InstanceConnectionManager.py
shubha-rajan: So this might be a dumb question, but if we pass the same timeout into the connector function that we use for fetching the metadata, won't that double the timeout? Like we could wait 25 sec to get metadata and another 25 to get the connection if the timeout was 30?

shubha-rajan: Would a better approach be to timeout for the whole function, instead of a timeout for just the call to `.result()`?

kurtisvg (changes requested):
Discussed some changes in handling asyncio/thread API differences offline as well.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -182,13 +181,10 @@ def __init__(
...
         self._priv_key, pub_key = generate_keys()
         self._pub_key = pub_key.decode("UTF-8")
-        self._lock = threading.Lock()
 
         logger.debug("Updating instance data")
 
-        with self._lock:
-            self._current = self._perform_refresh()
-            self._next = self.immediate_future(self._current)
+        asyncio.run_coroutine_threadsafe(self._perform_refresh(), self._loop).result()
```
kurtisvg, line 187: So. we don't want to do this, because this means that we're blocking on the refresh for the initialization to complete. 

Does this work? 
```
self._current = self._perform_refresh()
self._next = self._current
asyncio.run_coroutine_threadsafe(self._current, self._loop)
```

kurtisvg (changes requested):

google/cloud/sql/connector/connector.py
```diff
@@ -38,7 +38,7 @@ def _get_loop() -> asyncio.AbstractEventLoop:
     return _loop
 
 
-def connect(instance_connection_string, driver: str, **kwargs):
+def connect(instance_connection_string, driver: str, timeout: int = 30, **kwargs):
```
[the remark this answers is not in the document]
shubha-rajan: Good catch. `pymysql` accepts a `read_timeout`, `write_timeout` and `connect_timeout` so we're good there, but `pg8000` has a `timeout` arg. We should probably rename this to something that won't conflict. Although, I'm not sure if there would be a case where someone would want to pass in a general timeout directly to the driver and not to the connector.

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: Fixed

google/cloud/sql/connector/connector.py
```diff
@@ -85,4 +85,11 @@ def connect(instance_connection_string, driver: str, **kwargs):
         icm = InstanceConnectionManager(instance_connection_string, driver, keys, loop)
         _instances[instance_connection_string] = icm
 
-    return icm.connect(driver, user=kwargs.pop("user"), **kwargs)
+    if "connect_timeout" in kwargs:
+        timeout = kwargs.pop("connect_timeout")
+    elif "timeout" in kwargs:
+        timeout = kwargs.pop("timeout")
+    else:
+        timeout = 30
```
kurtisvg, line 93:
```suggestion
        timeout = 30 # 30s
```

kurtisvg (approved):

PR #66, opened 2021-03-26
chore: move static methods to their own modules

### LINKED ISSUE #61: InstanceConnectionManager.py should be "instance_connection_manager.py"

### LINKED ISSUE #62: Move static functions into private module functions
We have a couple of static methods being used in the ICM class that would probably be better as just standalone private functions. It should be possible to just pull out the refresh functionality into it's own file. Examples:

get_metadata: https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/851d1109d8f79f7e0a362aeb97d0512d7f162aa6/google/cloud/sql/connector/InstanceConnectionManager.py#L212-L218

get_ephemeral: https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/851d1109d8f79f7e0a362aeb97d0512d7f162aa6/google/cloud/sql/connector/InstanceConnectionManager.py#L281-L288

shubha-rajan:
Fixes #62


google/cloud/sql/connector/ephemeral.py
```diff
@@ -0,0 +1,91 @@
+"""
```
kurtisvg, line 1: nit: These functions probably don't each need their own file, they could probably be a single file (called `refresh` or `refresh_utils` maybe?)

google/cloud/sql/connector/ephemeral.py
```diff
@@ -0,0 +1,91 @@
...
+
+    logger.debug("Requesting ephemeral certificate")
+
+    if (
+        not isinstance(credentials, Credentials)
+        or not isinstance(project, str)
+        or not isinstance(instance, str)
+        or not isinstance(pub_key, str)
+    ):
+        raise TypeError("Cannot take None as an argument.")
```
kurtisvg, line 69: It would be nice if this was a bit more descriptive as to what went wrong or which failed

google/cloud/sql/connector/ephemeral.py
```diff
@@ -0,0 +1,91 @@
...
+
+    if not credentials.valid:
+        request = google.auth.transport.requests.Request()
+        credentials.refresh(request)
+
+    headers = {
+        "Authorization": "Bearer {}".format(credentials.token),
```
kurtisvg, line 76: nit: f-strings?

kurtisvg (approved):
Left some comments, but feel free to handle them in another PR if it's more convient

google/cloud/sql/connector/ephemeral.py
shubha-rajan: we can't do f-strings if we want to support python 3.6 :/

shubha-rajan: nvm turns out it wasn't that change which is causing the python 3.6 tests to fail. Now I'm trying to figure out what the actual problem is.

PR #69, opened 2021-04-01
feat: reuse instance connection managers

### LINKED ISSUE #64: ICMs should be reused on connect
ICM should be reused on connect calls instead of generating a different one each time connect is called

shubha-rajan:
fixes #64

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -28,6 +28,7 @@ _loop: Optional[asyncio.AbstractEventLoop] = None
 
 _instances = {}
 
+
 def _get_loop() -> asyncio.AbstractEventLoop:
     global _loop
     if _loop is None:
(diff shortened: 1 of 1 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -36,12 +37,12 @@ def _get_loop() -> asyncio.AbstractEventLoop:
...
-def connect(instance_connection_string, driver: str, **kwargs):
+def connect(instance_conn_str, driver: str, **kwargs):
     """Prepares and returns a database connection object and starts a
     background thread to refresh the certificates and metadata.
 
-    :type instance_connection_string: str
-    :param instance_connection_string:
+    :type instance_conn_str: str
+    :param instance_conn_str:
         A string containing the GCP project name, region name, and instance
```
kurtisvg, line 46: If you're going to shorten this, update the docstring to include "instance connection string" and an explanation of the format, since that's a significant term in helping users find it in the console.

google/cloud/sql/connector/connector.py
```diff
@@ -36,12 +37,12 @@ def _get_loop() -> asyncio.AbstractEventLoop:
     return _loop
 
 
-def connect(instance_connection_string, driver: str, **kwargs):
+def connect(instance_conn_str, driver: str, **kwargs):
```
kurtisvg, line 40: nit: if you're gonna shorten, shorten consistently 
```suggestion
def connect(inst_conn_str, driver: str, **kwargs):
```

kurtisvg (approved):

Issue #75, opened 2021-04-08
Python 3.6 build is flaky/timing out


shubha-rajan:
Logs from continuous build: https://fusion2.corp.google.com/invocations/63471ec3-4bec-45ea-a83b-8864e4d7a349

PR #76, opened 2021-04-08
fix: cancel all async resources when an ICM is destroyed

shubha-rajan:
I noticed the following errors in the logs for #75 so this might fix it.

- The correct class name is [`asyncio.CancelledError`](https://docs.python.org/3.8/library/asyncio-exceptions.html#asyncio.CancelledError) 

- The connection timeout test was missing a call to `del icm` and `timeout_stub` didn't return after timing out

- `connect_future` should be cancelled if it times out 

- `___del___` should await the cancellation of tasks after cancelling them


google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -189,17 +190,21 @@ def __del__(self):
...
+        async def _deconstruct():
+            if not self._client_session.closed:
+                logger.debug("Waiting for _client_session to close")
+                await self._client_session.close()
+            if self._current is not None:
+                logger.debug("Waiting for _current to be cancelled")
+                await complete_task_cancellation(self._current)
+            if self._next is not None:
+                logger.debug("Waiting for _next to be cancelled")
+                await complete_task_cancellation(self._next)
```
kurtisvg, line 202: It's probably better to use `asyncio.gather` to await all three, so they can be actually async

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -189,17 +190,21 @@ def __del__(self):
...
+
+        deconstruct_future = asyncio.run_coroutine_threadsafe(
+            _deconstruct(), loop=self._loop
+        )
+        deconstruct_future.result()
```
kurtisvg, line 207: What if this doesn't finish?

tests/unit/test_connector.py
```diff
@@ -41,3 +44,5 @@ async def timeout_stub(*args, **kwargs):
         pytest.raises(
             TimeoutError, connector.connect, connect_string, "pymysql", timeout=timeout
         )
+
+    del icm
```
kurtisvg, line 48: I'm not sure I understand why this deletion is necessary

google/cloud/sql/connector/instance_connection_manager.py
kurtisvg, line 207: We might want to handle this in a different PR, but I think we may have inadvertently started using an anti-pattern here and are forcing all of the `icm` close synchronously. 

What might be better is if we switch to having an object that holds all the references to the individual connection managers and closes them out in async, all at once.

kurtisvg (changes requested):

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: I'll open another issue and PR for this then and just remove the references to __del__ in this PR

tests/unit/test_connector.py
shubha-rajan: Removed this, as well as the other calls to del. There's no need to call it explicitly.

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: So it turns out that the `__del__` function was what was causing the tests to hang, so I added these changes back in, but I agree that we should close the connection managers out in async rather than sync calls for each one.

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: done

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: added a 5 second timeout. It's not the best solution, but it'll prevent it from blocking indefinitely. Ideally we'll remove this sync code when we refactor to resolve #77

kurtisvg: #77

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -189,17 +190,24 @@ def __del__(self):
...
-            close_future = asyncio.run_coroutine_threadsafe(
-                self.__client_session.close(), loop=self._loop
-            )
-            close_future.result()
-
+        async def _deconstruct():
+            tasks = []
+            if not self._client_session.closed:
+                logger.debug("Waiting for _client_session to close")
+                tasks.append(self._client_session.close)
```
kurtisvg, line 197: You still need to call these: 
```suggestion
                tasks.append(self._client_session.close())
```

Remember, calling an async function returns the coro object, when then has to be awaited to get the actual results. 

asyncio.gather does the waiting part for you.

google/cloud/sql/connector/utils.py
```diff
@@ -57,6 +57,15 @@ async def generate_keys():
     return priv_key, pub_key
 
 
+async def await_task_cancellation(task: asyncio.Task) -> None:
+    """A helper coroutine to await the cancellation of a task."""
+    task.cancel()
+    try:
+        await task
```
kurtisvg, line 64: Just to confirm -  are we sure we need to actually await the results? What happens if we just cancel?

kurtisvg (changes requested):

google/cloud/sql/connector/utils.py
shubha-rajan: awaiting the task guarantees that it was actually cancelled, since tasks can swallow the `CancelledError` and continue to run. https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: I updated this to use `asyncio.gather(*[task() for task in tasks])`

kurtisvg: Any reason you decided to defer the function call? postponing it feels less clear to me

google/cloud/sql/connector/utils.py
kurtisvg: It doesn't guarantee it was cancelled, just that the task was resolved:
 
> This arranges for a CancelledError exception to be thrown into the wrapped coroutine on the next cycle of the event loop.

This means that the CancelledError will get potentially thrown any time there is an "await" statement

> Therefore, unlike Future.cancel(), Task.cancel() does not guarantee that the Task will be cancelled, although suppressing cancellation completely is not common and is actively discouraged.

So you can cancel a task, but if if captures the error and suppresses, it just won't be canceled. 

So I don't think we need to await the result, because we're only executing one coro at a time (which means the task will be canceled on next context switch to it) and none of the functions we are calling are capturing the interruptions. If they were, we don't really care because we don't need the result anyway.

kurtisvg (approved):

PR #79, opened 2021-04-15
feat: allow specifying ip address type

### LINKED ISSUE #16: Finalize design for configuration arguments passed into connect()

shubha-rajan:
fixes #16


google/cloud/sql/connector/connector.py
```diff
@@ -48,7 +49,12 @@ def _get_keys(loop: asyncio.AbstractEventLoop) -> concurrent.futures.Future:
     return _keys
 
 
-def connect(instance_connection_string: str, driver: str, **kwargs: Any) -> Any:
+def connect(
+    instance_connection_string: str,
+    driver: str,
+    ip_types: List[IPTypes] = DEFAULT_IP_TYPES,
```
kurtisvg, line 55: Should we add some documentation for this arg?

tests/system/test_pg8000_connection.py
```diff
@@ -30,6 +31,7 @@ def getconn() -> pg8000.dbapi.Connection:
         conn: pg8000.dbapi.Connection = connector.connect(
             os.environ["POSTGRES_CONNECTION_NAME"],
             "pg8000",
+            ip_types=DEFAULT_IP_TYPES,
```
kurtisvg, line 34: This shouldn't need to be passed right? Since it has a default value?

tests/system/test_pymysql_connection.py
```diff
@@ -30,6 +31,7 @@ def getconn() -> pymysql.connections.Connection:
         conn: pymysql.connections.Connection = connector.connect(
             os.environ["MYSQL_CONNECTION_NAME"],
             "pymysql",
+            ip_types=DEFAULT_IP_TYPES,
```
kurtisvg, line 34: same here

tests/unit/test_connector.py
```diff
@@ -45,5 +46,10 @@ async def timeout_stub(*args: Any, **kwargs: Any) -> None:
     mock_instances[connect_string] = icm
     with patch.dict(connector._instances, mock_instances):
         pytest.raises(
-            TimeoutError, connector.connect, connect_string, "pymysql", timeout=timeout
+            TimeoutError,
+            connector.connect,
+            connect_string,
+            "pymysql",
+            ip_types=DEFAULT_IP_TYPES,
```
kurtisvg, line 53: and here

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -102,15 +123,17 @@ def __init__(
...
-    def __init__(self, *args: Any) -> None:
-        super(CloudSQLConnectionError, self).__init__(self, *args)
+    def get_preferred_ip(self, ip_types: List[IPTypes]) -> str:
+        """Returns the first IP address for the instance, in order of the
+        preference supplied by ip_types. If no IP addressess with the given
+        preference are found, an error is raised."""
+        for ip_type in ip_types:
+            if ip_type.value in self.ip_addrs:
+                return self.ip_addrs[ip_type.value]
+        raise CloudSQLConnectionError(
```
kurtisvg, line 133: The docstring for this error says:
```
Raised when the provided connection string is not formatted correctly.
```

We can probably address this in a different PR, but we probably need more descriptive error types.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -75,11 +96,11 @@ class InstanceMetadata:
     def __init__(
         self,
         ephemeral_cert: str,
-        ip_address: str,
+        ip_addrs: Dict[str, Any],
```
kurtisvg, line 99: In the new version of the Go connector, I've made the default "public" and required them to opt into private to help make it more of an explicit choice. What do you think about doing the same here?

https://github.com/kurtisvg/cloud-sql-go-connector/blob/925d6c2686d519d182dc196c752ed0c7edb0e28c/dialer.go#L93

kurtisvg (changes requested):
Can we also add a test with the different IP types? One with public and one with private? (although we may need to flag the private ip one)

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: I agree, let's keep them consistent. We should probably also make the same update in the Java connector: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/5ceacadbca5d91018b53a8131f3b5b110d70155d/core/src/main/java/com/google/cloud/sql/core/CoreSocketFactory.java#L78

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: I can add a different error type for this. I also just opened up a new issue to add more error types and better error messages.

google/cloud/sql/connector/instance_connection_manager.py
kurtisvg: This is probably a breaking change in the Java connector, but I think we should introduce it whenever we decide to next do a major version bump

kurtisvg (changes requested):

google/cloud/sql/connector/connector.py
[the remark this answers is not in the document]
shubha-rajan: I'll defer to your judgement on this. I know that we originally had the list so that users could set a fallback option ("if private is unavailable, connect to public" but I'm not sure if there's a real-world use case for that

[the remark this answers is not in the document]
kurtisvg: IMO, it's confusing because we only "fall back" when the IP doesn't exist - not when it's unavailable. If an instance has a private IP and a public IP, and they specify "PRIVATE,PUBLIC" it'll always use the Private IP. But I'm open to your thoughts on it.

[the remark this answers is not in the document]
shubha-rajan: in that case, I agree with you, let's simplify it. I doubt anyone using this library wouldn't know whether the IP address they want to use exists, and if it doesn't exist and they expect it to, they should get error feedback right away

google/cloud/sql/connector/connector.py
```diff
@@ -48,7 +49,12 @@ def _get_keys(loop: asyncio.AbstractEventLoop) -> concurrent.futures.Future:
     return _keys
 
 
-def connect(instance_connection_string: str, driver: str, **kwargs: Any) -> Any:
+def connect(
+    instance_connection_string: str,
+    driver: str,
+    ip_types: IPTypes = DEFAULT_IP_TYPE,
```
kurtisvg, line 55: I think DEFAULT_IP_TYPE is only used here, so we can just hardcode this to Public.

kurtisvg (approved):

PR #101, opened 2021-05-12
feat: add support for IAM auth with pg8000 driver

### LINKED ISSUE #100: Add support for Postgres IAM Authentication

shubha-rajan:
Fixes #100

### DIFF
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -54,16 +54,15 @@ logger = logging.getLogger(name=__name__)
 APPLICATION_NAME = "cloud-sql-python-connector"
 SERVER_PROXY_PORT = 3307
 
-# default_refresh_buffer is the minimum amount of time for which a
-# certificate must be valid to ensure the next refresh attempt has adequate
-# time to complete.
-_default_refresh_buffer: int = 5 * 60
+# default_refresh_buffer is the amount of time before a refresh's result expires
+# that a new refresh operation begins.
+_default_refresh_buffer: int = 5 * 60  # 5 minutes
 
-# iamAuthRefreshBuffer is the minimum amount of time for which a
-# certificate holding an Access Token must be valid. Because some token
-# sources are refreshed with only ~60 seconds before expiration, this value
-# must be smaller than the defaultRefreshBuffer.
-_iam_auth_refresh_buffer: int = 55
+# _iam_auth_refresh_buffer is the amount of time before a refresh's result expires
+# that a new refresh operation begins when IAM DB AuthN is enabled. Because token
+# sources may be cached until ~60 seconds before expiration, this value must be smaller
+# than default_refresh_buffer.
+_iam_auth_refresh_buffer: int = 55  # seconds
 
 
 class IPTypes(Enum):
@@ -113,7 +112,7 @@ class InstanceMetadata:
         private_key: bytes,
         server_ca_cert: str,
         expiration: datetime.datetime,
-        seconds_until_refresh: int
+        seconds_until_refresh: int,
     ) -> None:
         self.ip_addrs = ip_addrs
         self.context = ConnectionSSLContext()
@@ -297,20 +296,21 @@ class InstanceConnectionManager:
         expiration = datetime.datetime.strptime(
             x509.get_notAfter().decode("ascii"), "%Y%m%d%H%M%SZ"
         )
-        token_expiration: datetime.datetime = self._credentials.expiry
+        if self._credentials is not None:
+            token_expiration: datetime.datetime = self._credentials.expiry
 
         if expiration > token_expiration:
             expiration = token_expiration
 
-        seconds_until_refresh = self.seconds_until_refresh(expiration)
+        seconds_until_refresh = await self.seconds_until_refresh(expiration)
 
         return InstanceMetadata(
             ephemeral_cert,
             metadata["ip_addresses"],
             priv_key,
             metadata["server_ca_cert"],
-            expiration, 
-            seconds_until_refresh
+            expiration,
+            seconds_until_refresh,
         )
 
     def _auth_init(self) -> None:
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -23,10 +23,12 @@ from google.cloud.sql.connector.version import __version__ as version
 import asyncio
 import aiohttp
 import concurrent
+import datetime
 from enum import Enum
 import google.auth
 from google.auth.credentials import Credentials
 import google.auth.transport.requests
+import OpenSSL
 import ssl
 import socket
 from tempfile import TemporaryDirectory
@@ -52,9 +54,16 @@ logger = logging.getLogger(name=__name__)
 APPLICATION_NAME = "cloud-sql-python-connector"
 SERVER_PROXY_PORT = 3307
 
-# The default delay is set to 55 minutes since each ephemeral certificate is only
-# valid for an hour. This gives five minutes of buffer time.
-_delay: int = 55 * 60
+# default_refresh_buffer is the minimum amount of time for which a
+# certificate must be valid to ensure the next refresh attempt has adequate
+# time to complete.
+_default_refresh_buffer: int = 5 * 60
+
+# iamAuthRefreshBuffer is the minimum amount of time for which a
+# certificate holding an Access Token must be valid. Because some token
+# sources are refreshed with only ~60 seconds before expiration, this value
+# must be smaller than the defaultRefreshBuffer.
+_iam_auth_refresh_buffer: int = 55
 
 
 class IPTypes(Enum):
@@ -94,6 +103,7 @@ class CloudSQLIPTypeError(Exception):
 class InstanceMetadata:
     ip_addrs: Dict[str, Any]
     context: ssl.SSLContext
+    cert_expiration: datetime.datetime
 
     def __init__(
         self,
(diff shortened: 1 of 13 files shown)


google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -52,9 +54,16 @@
 APPLICATION_NAME = "cloud-sql-python-connector"
 SERVER_PROXY_PORT = 3307
 
-# The default delay is set to 55 minutes since each ephemeral certificate is only
-# valid for an hour. This gives five minutes of buffer time.
-_delay: int = 55 * 60
+# default_refresh_buffer is the minimum amount of time for which a
+# certificate must be valid to ensure the next refresh attempt has adequate
+# time to complete.
+_default_refresh_buffer: int = 5 * 60
```
kurtisvg, line 60:
```suggestion
_default_refresh_buffer: int = 5 * 60 # 5 minutes
```

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -52,9 +54,16 @@
 APPLICATION_NAME = "cloud-sql-python-connector"
 SERVER_PROXY_PORT = 3307
 
-# The default delay is set to 55 minutes since each ephemeral certificate is only
-# valid for an hour. This gives five minutes of buffer time.
-_delay: int = 55 * 60
+# default_refresh_buffer is the minimum amount of time for which a
+# certificate must be valid to ensure the next refresh attempt has adequate
+# time to complete.
```
kurtisvg, line 59:
```suggestion
# default_refresh_buffer is the amount of time before a refresh's result expires
# that a new refresh operation begins. 
```

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -52,9 +54,16 @@
...
+_default_refresh_buffer: int = 5 * 60
+
+# iamAuthRefreshBuffer is the minimum amount of time for which a
+# certificate holding an Access Token must be valid. Because some token
+# sources are refreshed with only ~60 seconds before expiration, this value
+# must be smaller than the defaultRefreshBuffer.
```
kurtisvg, line 65:
```suggestion
# _iam_auth_refresh_buffer is the amount of time before a refresh's result expires
# that a new refresh operation begins when IAM DB AuthN is enabled. Because token 
# sources may be cached until ~60 seconds before expiration, this value must be smaller
# than default_refresh_buffer.
```

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -52,9 +54,16 @@
...
+_iam_auth_refresh_buffer: int = 55
```
kurtisvg, line 66:
```suggestion
_iam_auth_refresh_buffer: int = 55 # seconds
```

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -287,6 +314,32 @@ def _auth_init(self) -> None:
 
         self._credentials = credentials
 
+    async def seconds_until_refresh(self) -> int:
```
kurtisvg, line 317: Same as in Java - I think this can be a field that is set before the refresh's result is returned. It should be both calculated once (per result), and specific to a refresh result.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -299,22 +352,21 @@ async def _perform_refresh(self) -> asyncio.Task:
 
         self._current = self._loop.create_task(self._get_instance_data())
         # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
-        self._next = self._loop.create_task(self._schedule_refresh(_delay))
+        self._next = self._loop.create_task(self._schedule_refresh())
 
         return self._current
 
-    async def _schedule_refresh(self, delay: int) -> asyncio.Task:
+    async def _schedule_refresh(self) -> asyncio.Task:
```
kurtisvg, line 359: IMO, we should leave this in. Both the Java connector and the Go connector have a way to force a refresh early if something goes wrong, we should consider doing so here.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -94,6 +103,7 @@ def __init__(self, *args: Any) -> None:
 class InstanceMetadata:
     ip_addrs: Dict[str, Any]
     context: ssl.SSLContext
+    cert_expiration: datetime.datetime
```
kurtisvg, line 106: nit: Maybe just `expiration`? It should be the expiration of the data as a whole (both token/cert)

kurtisvg (changes requested):

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: fixed

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: fixed

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: This is now being called in `get_instance_data` and is passed in as an attribute of `InstanceData`

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -94,16 +102,22 @@ def __init__(self, *args: Any) -> None:
 class InstanceMetadata:
     ip_addrs: Dict[str, Any]
     context: ssl.SSLContext
+    expiration: datetime.datetime
+    seconds_until_refresh: int
```
kurtisvg, line 106: Hmm, why are we storing this value on the object? Won't it a) become invalid almost instantly and b) only be used once to schedule the next refresh?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -299,22 +358,22 @@ async def _perform_refresh(self) -> asyncio.Task:
 
         self._current = self._loop.create_task(self._get_instance_data())
         # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
-        self._next = self._loop.create_task(self._schedule_refresh(_delay))
+        self._next = self._loop.create_task(self._schedule_refresh())
 
         return self._current
 
-    async def _schedule_refresh(self, delay: int) -> asyncio.Task:
+    async def _schedule_refresh(self, delay: int = None) -> asyncio.Task:
```
kurtisvg, line 365: should this be 0? (or Optional[int])?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -287,6 +327,25 @@ def _auth_init(self) -> None:
 
         self._credentials = credentials
 
+    async def seconds_until_refresh(self, expiration: datetime.datetime) -> int:
```
kurtisvg, line 330: Does this need to be async? I don't see any awaits happening inside it.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -299,22 +358,22 @@ async def _perform_refresh(self) -> asyncio.Task:
 
         self._current = self._loop.create_task(self._get_instance_data())
         # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
-        self._next = self._loop.create_task(self._schedule_refresh(_delay))
+        self._next = self._loop.create_task(self._schedule_refresh())
```
kurtisvg, line 361: Can we move the delay calculation into this section? 

I feel like I'm becoming a bit disjointed between how this operates and how the java version operates. This method doesn't seem to actually be performing any refresh, just scheduling it. The schedule refresh doesn't seem to be scheduling, just performing. Is this confusing for you as well?

kurtisvg (changes requested):

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: Originally we were awaiting the metadata in the async function to calculate the expiration, but then that was moved so we calculate the seconds_until_refresh when creating the metadata. 
Now we're awaiting the metatata in this function again so it needs to be async.

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: So I've tried moving the delay calculations into this section, but I get this test failure:
```
    async def seconds_until_refresh(self) -> int:
>       expiration = (await self._current).expiration
E       RuntimeError: Task <Task pending coro=<test_InstanceConnectionManager_perform_refresh() running at /Users/shubhadayini/cloud-sql-python-connector/tests/unit/test_instance_connection_manager.py:63> cb=[_run_until_complete_cb() at /Users/shubhadayini/.pyenv/versions/3.7.6/lib/python3.7/asyncio/base_events.py:157]> got Future <Task pending coro=<InstanceConnectionManager._get_instance_data() running at /Users/shubhadayini/cloud-sql-python-connector/google/cloud/sql/connector/instance_connection_manager.py:259>> attached to a different loop
```

It might also be an issue with how the test is written. Calling the delay calculation in schedule refresh makes the error go away.

shubha-rajan: from what I understand schedule_refresh sleeps for `delay` seconds before performing the next refresh. it's not as clean as the executor.submit in Java though

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: changed to `Optional[int]`

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: You're right, it doesn't really make sense. I think I just wanted to do all the calculations while creating the ephemeral cert. Updated to only store the expiration

kurtisvg (approved):

PR #105, opened 2021-05-17
chore: Add better connect instructions and IAM DB authN to README

shubha-rajan:


README.md
```diff
@@ -53,30 +53,57 @@ from google.cloud.sql.connector import connector
...
     "pymysql",
     ip_types=IPTypes.PRIVATE # Prefer private IP
 ... insert other kwargs ...
 )
 ```
 
 Note: If specifying Private IP, your application must already be in the same VPC network as your Cloud SQL Instance. 
+
+### IAM Authentication
+Connections using [IAM database authentication](https://cloud.google.com/sql/docs/postgres/iam-logins) are supported when using the Postgres driver. First, make sure to [configure your Cloud SQL Instance to allow IAM authentication](https://cloud.google.com/sql/docs/postgres/create-edit-iam-instances#configure-iam-db-instance) and [add an IAM database user](https://cloud.google.com/sql/docs/postgres/create-manage-iam-users#creating-a-database-user). Then, you can connect using user or service account credentials instead of a password when the `enable_iam_auth` keyword argument is set to true, and `user` is set to the email address associated with the IAM user.
```
enocom, line 96: Minor wording nit. 

For the sentence that starts with "Then, you ...", how about the following?

Now, you can connect using user or service account credentials instead of a password. In the call to `connect`, set the `enable_iam_auth` keyword argument to true and `user` to the email address associated with your IAM user.

Also, should we wrap this one line across a few lines?

kurtisvg (approved):

PR #116, opened 2021-06-14
chore: add note about Postgres-only IAM DB Authn

enocom:


enocom:
@shubha-rajan Are there other places to note that IAM DB Authn is Postgres only?

shubha-rajan:
> @shubha-rajan Are there other places to note that IAM DB Authn is Postgres only?

Yes, I think specifically in [this docstring](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/connector.py#L77) where we document the `enable_iam_auth` parameter

enocom:
Thanks! Made a minor wording change to match the Proxy: https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/822/files.

kurtisvg (approved):

PR #117, opened 2021-06-14
chore: use "Postgres" not PostgreSQL in docs

enocom:


enocom:
Matches https://github.com/GoogleCloudPlatform/cloudsql-proxy/pull/822 and https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/523.

kurtisvg (approved):

PR #131, opened 2021-07-26
feat: support Active Directory auth for Windows

shubha-rajan:
Add support for Integrated Security (Active Directory auth) for Windows using [SSPI](https://docs.microsoft.com/en-us/windows-server/security/windows-authentication/security-support-provider-interface-architecture). PyTDS offers experimental support for Kerberos auth on non-Windows platforms, but I haven't tested it yet, so for now only Windows will be supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ connector.connect(
  )
 ```
 
-### SQL Server Active Directory Authentication (Integrated Security)
+### SQL Server Active Directory Authentication
 Active Directory authentication for SQL Server instances is currently only supported on Windows. First, make sure to follow [these steps](https://cloud.google.com/blog/topics/developers-practitioners/creating-sql-server-instance-integrated-active-directory-using-google-cloud-sql) to set up a Managed AD domain and join your Cloud SQL instance to the domain. [See here for more info on Cloud SQL Active Directory integration](https://cloud.google.com/sql/docs/sqlserver/ad).
 
 Once you have followed the steps linked above, you can run the following code to return a connection object:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -127,6 +127,32 @@ connector.connect(
...
+### SQL Server Active Directory Authentication (Integrated Security)
+Active Directory authentication for SQL Server instances is currently only supported on Windows. First, make sure to follow [these steps](https://cloud.google.com/blog/topics/developers-practitioners/creating-sql-server-instance-integrated-active-directory-using-google-cloud-sql) to set up a Managed AD domain and join your Cloud SQL instance to the domain. [See here for more info on Cloud SQL Active Directory integration](https://cloud.google.com/sql/docs/sqlserver/ad).
+
+Once you have followed the steps linked above, you can run the following code to return a connection object:
+```python
+connector.connect(
+    "project:region:instance",
+    "pytds",
+    db="my_database",
+    integrated_security=True,
```
kurtisvg, line 140: Do you think `integrated_security` is the best name here? Or would active-directory-auth be better? Which is more obvious to users what it's doing?

kurtisvg (approved):

README.md
shubha-rajan: I went with the parameter that the JDBC SQL Server driver used, but if active_directory_auth is better we can change it.

kurtisvg: I just want it to be easily discoverable. At a passing glance `integrated_security` doesn't make sense to me, but might be the magic words the AD folks are familiar with. 

I'll defer to your judgement on the issue.

PR #135, opened 2021-07-28
fix: only replace refresh result if successful or current is invalid

### LINKED ISSUE #134: Only replace refresh result if refresh was successful or current result is invalid
## Bug Description

Currently, the Java connector performs a refresh is always replaces the previous result with a new result. However sometimes, this result can fail and it replaces the current value. Instead, we should try to avoid replacing until it's complete.

See related issue for Java connector: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/550

shubha-rajan:
fixes #134

### DIFF
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -395,7 +395,7 @@ class InstanceConnectionManager:
                 try:
                     if isinstance(self._current, asyncio.Task):
                         # if _current is a completed asyncio.Task we check its result
-                        # and only replace it if it's invalid. 
+                        # and only replace it if it's invalid.
                         # Otherwise, if _current is a Future (first request), we always
                         # replace it
                         instance_data = self._current.result()
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -36,7 +36,6 @@ from tempfile import TemporaryDirectory
 from typing import (
     Any,
     Awaitable,
-    Coroutine,
     Dict,
     Optional,
     TYPE_CHECKING,
@@ -220,8 +219,8 @@ class InstanceConnectionManager:
     _project: str
     _region: str
 
-    _current: Union[Coroutine, asyncio.Task]
-    _next: Union[Coroutine, asyncio.Task]
+    _current: Union[concurrent.futures.Future, asyncio.Task]
+    _next: Union[concurrent.futures.Future, asyncio.Task]
 
     def __init__(
         self,
@@ -254,9 +253,10 @@ class InstanceConnectionManager:
 
         logger.debug("Updating instance data")
 
-        self._current = self._perform_refresh()
+        self._current = asyncio.run_coroutine_threadsafe(
+            self._perform_refresh(), self._loop
+        )
         self._next = self._current
-        asyncio.run_coroutine_threadsafe(self._current, self._loop)
 
     def __del__(self) -> None:
         """Deconstructor to make sure ClientSession is closed and tasks have
@@ -351,7 +351,7 @@ class InstanceConnectionManager:
         self._credentials = credentials
 
     async def seconds_until_refresh(self) -> int:
-        expiration = (await self._current).expiration
+        expiration = self._current.result().expiration
 
         if self._enable_iam_auth:
             refresh_buffer = _iam_auth_refresh_buffer
@@ -393,7 +393,8 @@ class InstanceConnectionManager:
                 )
                 instance_data = None
                 try:
-                    instance_data = self._current.result()
+                    if isinstance(self._current, asyncio.Task):
+                        instance_data = self._current.result()
                 except Exception:
                     # Current result is invalid, no-op
                     logger.debug("Current instance data is invalid.")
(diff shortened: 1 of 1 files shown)


google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -210,8 +209,8 @@ def _client_session(self) -> aiohttp.ClientSession:
     _project: str
     _region: str
 
-    _current: Union[Coroutine, asyncio.Task]
-    _next: Union[Coroutine, asyncio.Task]
+    _current: Union[concurrent.futures.Future, asyncio.Task]
```
kurtisvg, line 212: I thought we weren't using futures any more?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -371,12 +371,40 @@ async def _perform_refresh(self) -> asyncio.Task:
...
-        self._next = self._loop.create_task(self._schedule_refresh())
-
-        return self._current
+        refresh_task = self._loop.create_task(self._get_instance_data())
+
+        async def _refresh_callback(task: asyncio.Task) -> None:
+            try:
+                task.result()
+            except Exception as e:
+                logger.warn(
```
kurtisvg, line 380: Do we really want to log this, or would it be better to include the context in the Exception?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -371,12 +371,40 @@ async def _perform_refresh(self) -> asyncio.Task:
...
+                    "An error occurred while performing refresh. Retrying immediately.",
+                    e,
+                )
+                instance_data = None
+                try:
+                    if isinstance(self._current, asyncio.Task):
```
kurtisvg, line 386: What's going on here? Maybe a comment would be good.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -371,12 +371,40 @@ async def _perform_refresh(self) -> asyncio.Task:
...
+                except Exception:
+                    # Current result is invalid, no-op
+                    logger.debug("Current instance data is invalid.")
+                if (
+                    instance_data is None
+                    or instance_data.expiration < datetime.datetime.now()
+                ):
+                    self._current = refresh_task
+                    # TODO: Implement force refresh method and a rate-limiter for perform_refresh
+                    # Retry by scheduling a refresh 60s from now.
```
kurtisvg, line 397: Maybe we should make this PR first

kurtisvg (changes requested):
I'm confused by the pivot back to futures.

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: So the initial perform_refresh has to be kicked off from another thread using `run_coroutine_threadsafe`. When I had current set to a coroutine and tried to await it, I got an error that said `RuntimeError: cannot reuse already awaited coroutine`. If we use a wrapped future just for that initial request, we can call `await` on it as many times as we want, and if it's already completed, we just get the result. I just updated the code again to store the perform_refresh coroutine again, but I think using a future just for that first request is a better approach

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: We're logging it in the Java Connector: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#L381-L382

Calling `task.result` on a completed task will either return the `InstanceData` object or throw an error. This `try/catch` block is the equivalent of defining the `onSuccess` and `onFailure` in the Java `FutureCallback`

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: Setting it to a coroutine gives this error: https://source.cloud.google.com/results/invocations/43cbe38e-181c-48d1-ba35-bea135868b7c/targets/cloud-devrel%2Fcloud-sql%2Fpython-connector%2Fpython3.7%2Flinux%2Fpresubmit/log

Just set it back to a future.

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: Added a comment

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: I actually think we should do this one first because the connector currently has no retry mechanism. If we get a failed refresh attempt, we're just stuck with it for the next hour :/

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -381,12 +381,39 @@ async def _perform_refresh(self) -> asyncio.Task:
...
+                try:
+                    instance_data = self._current.result()
+                except Exception:
+                    # Current result is invalid, no-op
+                    logger.debug("Current instance data is invalid.")
+                if (
+                    instance_data is None
+                    or instance_data.expiration < datetime.datetime.now()
+                ):
+                    self._current = refresh_task
```
kurtisvg, line 404: Should this be `task ` or `refresh_task`?

kurtisvg (approved):

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: either technically works since it's a closure, but task is probably better

PR #146, opened 2021-08-05
feat: add rate limiter and force refresh function

### LINKED ISSUE #136: Add a forceRefresh function that is called whenever a connection attempt fails
## Feature Description
Add a forceRefresh function that is called whenever a connection attempt fails. It is also necessary to implement a rate limiter on the `performRefresh` function so we don't exceed SQLAdmin API quota

shubha-rajan:
fixes #136 
Let me know if I should add more comments explaining the rate limiter implementation.

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -117,8 +117,6 @@ def connect(
         timeout = 30  # 30s
     try:
         return icm.connect(driver, ip_types, timeout, **kwargs)
-    except TimeoutError as e:
-        raise (e)  # if the error is because of a timeout, let it bubble up
     except Exception as e:
         # with any other exception, we attempt a force refresh, then throw the error
         icm.force_refresh()
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 import asyncio
 import concurrent
+import logging
 from google.cloud.sql.connector.instance_connection_manager import (
     InstanceConnectionManager,
     IPTypes,
@@ -30,6 +31,8 @@ _loop: Optional[asyncio.AbstractEventLoop] = None
 _keys: Optional[concurrent.futures.Future] = None
 
 _instances: Dict[str, InstanceConnectionManager] = {}
+
+logger = logging.getLogger(name=__name__)
 
 
 def _get_loop() -> asyncio.AbstractEventLoop:
@@ -112,5 +115,11 @@ def connect(
         timeout = kwargs["connect_timeout"]
     else:
         timeout = 30  # 30s
-
-    return icm.connect(driver, ip_types, timeout, **kwargs)
+    try:
+        return icm.connect(driver, ip_types, timeout, **kwargs)
+    except TimeoutError as e:
+        raise (e)  # if the error is because of a timeout, let it bubble up
+    except Exception as e:
+        # with any other exception, we attempt a force refresh, then throw the error
+        icm.force_refresh()
+        raise (e)
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -252,7 +252,8 @@ class InstanceConnectionManager:
         self._auth_init()
 
         self._refresh_rate_limiter = AsyncRateLimiter(
-            burst_size=1, interval=60, queue_size=1, loop=self._loop)
+            burst_size=1, interval=60, queue_size=1, loop=self._loop
+        )
 
         async def _set_instance_data() -> None:
             logger.debug("Updating instance data")
@@ -353,28 +354,24 @@ class InstanceConnectionManager:
         )
 
         self._credentials = credentials
-    
+
     async def _force_refresh(self) -> bool:
-            try:
-                self._current = await self._perform_refresh()
-                return True
-            except asyncio.queues.QueueFull: 
-                # a refresh attempt is already queued, so just block on the result
-                self._current = await self._next
-                return True
-            except Exception as e:
-                # if anything else goes wrong, log the error and return false
-                logger.exception(
-                    "Error occurred during force refresh attempt",
-                    exc_info=e
-                )
-                return False
+        try:
+            self._current = await self._perform_refresh()
+            return True
+        except asyncio.queues.QueueFull:
+            # a refresh attempt is already queued, so just block on the result
+            self._current = await self._next
+            return True
+        except Exception as e:
+            # if anything else goes wrong, log the error and return false
+            logger.exception("Error occurred during force refresh attempt", exc_info=e)
+            return False
 
-
-    def force_refresh(self, timeout: Optional[int]=None) -> bool:
-        return asyncio.run_coroutine_threadsafe(self._force_refresh, self._loop).result(timeout=timeout)
-        
-        
+    def force_refresh(self, timeout: Optional[int] = None) -> bool:
+        return asyncio.run_coroutine_threadsafe(
+            self._force_refresh(), self._loop
+        ).result(timeout=timeout)
 
     async def seconds_until_refresh(self) -> int:
         expiration = (await self._current).expiration
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -253,7 +253,7 @@ class InstanceConnectionManager:
         self._auth_init()
 
         self._refresh_rate_limiter = AsyncRateLimiter(
-            burst_size=2, interval=60, loop=self._loop
+            max_capacity=2, rate=1 / 30, loop=self._loop
         )
 
         async def _set_instance_data() -> None:
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -370,9 +370,7 @@ class InstanceConnectionManager:
             return True
         except Exception as e:
             # if anything else goes wrong, log the error and return false
-            logger.exception(
-                "Error occurred during force refresh attempt", exc_info=e
-            )
+            logger.exception("Error occurred during force refresh attempt", exc_info=e)
             return False
 
     def force_refresh(self, timeout: Optional[int] = None) -> bool:
@@ -451,7 +449,7 @@ class InstanceConnectionManager:
         :rtype: asyncio.Task
         :returns: A Task representing _get_instance_data.
         """
-                
+
         if delay is None:
             delay = await self.seconds_until_refresh()
         try:
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 # Custom utils import
+from google.cloud.sql.connector.rate_limiter import AsyncRateLimiter
 from google.cloud.sql.connector.refresh_utils import _get_ephemeral, _get_metadata
 from google.cloud.sql.connector.utils import write_to_file
 from google.cloud.sql.connector.version import __version__ as version
@@ -250,6 +251,9 @@ class InstanceConnectionManager:
         self._keys = asyncio.wrap_future(keys, loop=self._loop)
         self._auth_init()
 
+        self._refresh_rate_limiter = AsyncRateLimiter(
+            burst_size=1, interval=60, queue_size=1, loop=self._loop)
+
         async def _set_instance_data() -> None:
             logger.debug("Updating instance data")
             self._current = self._loop.create_task(self._get_instance_data())
@@ -378,7 +382,7 @@ class InstanceConnectionManager:
         :rtype: concurrent.future.Futures
         :returns: A future representing the creation of an SSLcontext.
         """
-
+        await self._refresh_rate_limiter.acquire()
         logger.debug("Entered _perform_refresh")
 
         refresh_task = self._loop.create_task(self._get_instance_data())
@@ -387,7 +391,8 @@ class InstanceConnectionManager:
             await refresh_task
         except Exception as e:
             logger.exception(
-                "An error occurred while performing refresh. Retrying in 60s.",
+                "An error occurred while performing refresh."
+                "Scheduling another refresh attempt immediately",
                 exc_info=e,
             )
             instance_data = None
@@ -401,9 +406,7 @@ class InstanceConnectionManager:
                 or instance_data.expiration < datetime.datetime.now()
             ):
                 self._current = refresh_task
-                # TODO: Implement force refresh method and a rate-limiter for perform_refresh
-                # Retry by scheduling a refresh 60s from now.
-                self._next = self._loop.create_task(self._schedule_refresh(60))
+                self._next = self._loop.create_task(self._perform_refresh())
 
         else:
             self._current = refresh_task
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -220,8 +220,8 @@ class InstanceConnectionManager:
     _region: str
 
     _refresh_in_progress: asyncio.locks.Event
-    _current: asyncio.Task
-    _next: asyncio.Task
+    _current: asyncio.Task[InstanceMetadata]
+    _next: asyncio.Task[asyncio.Task[InstanceMetadata]]
 
     def __init__(
         self,
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -252,7 +252,7 @@ class InstanceConnectionManager:
         self._auth_init()
 
         self._refresh_rate_limiter = AsyncRateLimiter(
-            burst_size=1, interval=60, queue_size=1, loop=self._loop
+            burst_size=2, interval=60, queue_size=2, loop=self._loop
         )
 
         async def _set_instance_data() -> None:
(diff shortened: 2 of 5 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -112,5 +115,11 @@ def connect(
         timeout = kwargs["connect_timeout"]
     else:
         timeout = 30  # 30s
-
-    return icm.connect(driver, ip_types, timeout, **kwargs)
+    try:
+        return icm.connect(driver, ip_types, timeout, **kwargs)
+    except TimeoutError as e:
+        raise (e)  # if the error is because of a timeout, let it bubble up
```
kurtisvg, line 121: Are we sure we want an exception for TimeoutError here? 

What if timeout is because the instance's IP has changed, and we need to refresh to get the updated IP?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -350,6 +355,24 @@ def _auth_init(self) -> None:
...
 
+    async def _force_refresh(self) -> bool:
+        try:
+            self._current = await self._perform_refresh()
+            return True
+        except asyncio.queues.QueueFull:
+            # a refresh attempt is already queued, so just block on the result
+            self._current = await self._next
+            return True
+        except Exception as e:
```
kurtisvg, line 366: This logic seems wrong when compared to the java connector: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#L307-L313

I think we probably want to mirror that behavior here?

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,126 @@
...
+
+class AsyncRateLimiter(object):
+    """
+    An asyncio-compatible rate limiter which uses the Leaky Bucket algorithm
+    (https://en.wikipedia.org/wiki/Leaky_bucket) to limit the number of function
+    calls over a time interval using an event queue.
+
+    :type burst_size: int
+    :param: burst_size:
+        The max number of function calls that can occur at the same time
```
kurtisvg, line 31: This is the bucket capacity right? So not necessarily the number of calls that can occur at the same time, but more so the max the bucket can fill up to when not being used?

kurtisvg: Ok, so this is my mistake - I actually gave you the wrong name of the algorithm. I said "Leaky Bucket", but I really meant "Token Bucket": https://en.wikipedia.org/wiki/Token_bucket

I think the token bucket is preferred because it'll let us "save up tokens" when we aren't using, and let us burst slightly if we need too. 

WDYT about using a token bucket instead?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -250,6 +251,10 @@ def __init__(
         self._keys = asyncio.wrap_future(keys, loop=self._loop)
         self._auth_init()
 
+        self._refresh_rate_limiter = AsyncRateLimiter(
+            burst_size=1, interval=60, queue_size=1, loop=self._loop
```
kurtisvg, line 255: I think we probably want to limit this at 2 per 60s (I think that's what Java and Go use)

kurtisvg (changes requested):

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: I actually had this using the token bucket algorithm originally (which is why those comments about bursts were left in, I forgot to remove them). I switched to leaky bucket because I thought it's better for our use case, since I figured we didn't want a bunch of refresh calls occurring at around the same time (i.e. a refresh starts before another one can complete). I'm not sure if bursting is useful for our use case, or if it might actually end up causing issues.

google/cloud/sql/connector/connector.py
shubha-rajan: That's a good point. Updated this to force a refresh on any error.

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: I updated it to use a token bucket again, but I'm still not convinced it's the better option.

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: Updated this to do 2 refreshes per 60s, but Java does 1/min. Should that be updated too? https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/5cfa39e49a5362babf8412256f0081cda21c414f/core/src/main/java/com/google/cloud/sql/core/CloudSqlInstance.java#L118

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: The code for python is going to look a little bit different because Python tasks don't behave like Java futures. 

https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel
Java futures [can reject cancellation if a task is in progress](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html#cancel(boolean)). When an asyncio task is cancelled, it throws a `CancelledError` in the wrapped coroutine. The coroutine can then either handle and suppress or throw that error. Since we're just wrapping `_perform_refresh` in a coroutine with a sleep (`_schedule_refresh`) to schedule the next refresh instead of an executor like in Java, the refresh will just get cancelled, regardless of whether it's sleeping or running.

The code above attempts another refresh, but if the rate limiter queue is full, we know there's a `_perform_refresh` call that's about to run. If schedule_refresh is still sleeping, then the queue will be empty. If the queue is full, we just block on the pending `_next` result, like in Java. We don't have to update `_next` ourselves, because awaiting `_perform_refresh()` will schedule the next task.

google/cloud/sql/connector/rate_limiter.py
kurtisvg: I think if we can figure out the other thread we have going on (how to cancel a not-yet-started task), that we won't need to worry about more than one task going at the same time. 

The reason I think token is preferred over leaky is because leaky limits outgoing traffic, but has a queue. We don't really ever need to queue attempts, we can stop once a perform refresh is successful. Token on the other hand, makes it OK to perform 2-3 refreshes in a row that helps weed out transient errors, but ensures that amortized over a couple of minutes, our consumption will work as expected.

google/cloud/sql/connector/instance_connection_manager.py
kurtisvg: Yeah, I think 2/min is probably an acceptable rate.

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: Could we switch to using one of [these methods](https://docs.python.org/3/library/asyncio-eventloop.html#scheduling-delayed-callbacks) instead of the `schedule_refresh` coroutine to schedule tasks? The `asyncio.TimerHandle` lets us schedule a callback at a specific time and then cancel it. So when scheduling the next refresh, we could do something like:
```
self._next = loop.call_later(self.seconds_until_refresh(), lambda: loop.create_task(self.perform_refresh()))
```

kurtisvg: That seems reasonable to me, as long as we can tell if `cancel` was successful or not.

shubha-rajan: I ended up going with the event approach that you suggested in the other thread. I also added a test to confirm that force_refresh works.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -419,17 +452,15 @@ async def _schedule_refresh(self, delay: Optional[int] = None) -> asyncio.Task:
         :rtype: asyncio.Task
         :returns: A Task representing _get_instance_data.
         """
-        logger.debug("Entering sleep")
 
+        logger.debug("Entering sleep")
```
kurtisvg, line 456: nit: should this be under the delay?

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,114 @@
...
+from types import TracebackType
+
+
+class AsyncRateLimiter(object):
+    """
+    An asyncio-compatible rate limiter which uses the Token Bucket algorithm
+    (https://en.wikipedia.org/wiki/Token_bucket) to limit the number of function calls over a time interval using an event queue.
+
+    :type burst_size: int
+    :param: burst_size:
```
kurtisvg, line 29: Should this be "the maximum capacity of the bucket will store at any one time"

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,114 @@
...
+        Default: 1
+
+    :type interval: int
+    :param: interval:
+        The period of time over which a number of calls equal to burst_size are allowed to complete. Default: 60s
+
+    :type queue_size: int
+    :param: interval:
+        The number of tasks that are allowed to be queued if rate limiter cannot be acquired. If exceeded, a QueueFull error will be thrown. A queue size of zero means the queue length has no upper bound. If setting queue size to a value greater than zero, queue_size must be greater than or equal to burst_size
+        Default: 0
```
kurtisvg, line 40: I don't see queue_size - should this doc string be updated

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def _add_tokens(self) -> None:
+        """
+        Calculates how much time has passed since the last leak and removes the
+        appropriate amount of events from the queue.
+        Leaking is done lazily, meaning that if there is a large time gap between
+        leaks, the next set of calls might be a burst if burst_size > 1
+        """
+        time_elapsed = time.time() - self._last_token_update
+        amount = math.floor(time_elapsed * self.burst_size / self.interval)
```
kurtisvg, line 69: nit: We may want to add parenthesis here, I can't remember if python respects the order of operations

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,114 @@
...
+        # don't allow more than burst_size to leak at once
+        leak_amount = min(amount, self.burst_size)
```
kurtisvg, line 71: hmm I don't think this is right... It shouldn't be the amount we leak, it should be the max capacity on hand

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,114 @@
...
+            except ValueError:
+                # we want to keep adding tokens until the semaphore reaches its upper bound
+                break
+
+    def _token_acquired(self, waiter_task: asyncio.Task) -> bool:
+        """
+        Releases tasks out of the queue if enough time has elapsed.
+        """
+        self._add_tokens()
+        return waiter_task.done()
```
kurtisvg, line 86: I'm not following what this is doing.

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,114 @@
...
+        await self.acquire()
+        return None
+
+    async def __aexit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc: Optional[BaseException],
+        tb: Optional[TracebackType],
+    ) -> None:
+        return None
```
kurtisvg, line 114: it's neat you added these here, but are you actually using them?

kurtisvg (changes requested):

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -350,6 +357,30 @@ def _auth_init(self) -> None:
 
         self._credentials = credentials
 
+    async def _force_refresh(self) -> bool:
+        if self._refresh_in_progress.is_set():
+            # if a refresh is in progress, then block on the result
+            self._current = await self._next
```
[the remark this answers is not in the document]
shubha-rajan: The Java Connector blocks on the result. Also, `_next` is a task that wraps the refresh task that's returned by `perform_refresh`. so we do need to await it to get the task that we're wrapping

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: I think the parentheses are unnecessary here. Python respects order of operations and even if it did, we're only multiplying and dividing so it doesn't matter. the result is the same regardless of which one you do first.

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: sorry forgot to change this when I updated to using token bucket. fixed in latest changes.

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: I'm using a the rate limiter as a context manager in the tests. I was using the context manager in `perform_refresh` in an earlier draft, but switched it to just using an acquire. I could delete these and change the tests if you want.

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: the waiter task is waiting to acquire the semaphore (i.e. acquire a token) . After adding tokens, this checks to see if the task has gotten a token, and if it has, it breaks out of the loop in `acquire`. I'm using the loop so that we can periodically check if enough time has passed to add another token.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -350,6 +357,30 @@ def _auth_init(self) -> None:
 
         self._credentials = credentials
 
+    async def _force_refresh(self) -> bool:
+        if self._refresh_in_progress.is_set():
+            # if a refresh is in progress, then block on the result
```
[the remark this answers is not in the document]
shubha-rajan: done

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -350,6 +357,30 @@ def _auth_init(self) -> None:
...
+            self._current = await self._next
+            return True
+        else:
```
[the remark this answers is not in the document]
shubha-rajan: done

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: fixed.

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: fixed

google/cloud/sql/connector/rate_limiter.py
shubha-rajan: fixed

google/cloud/sql/connector/rate_limiter.py
kurtisvg: yeah I think we should update the tests to use the same method.

google/cloud/sql/connector/instance_connection_manager.py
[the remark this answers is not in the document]
kurtisvg: We discussed making some changes to this in a separate PR.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -350,6 +357,27 @@ def _auth_init(self) -> None:
...
+            # schedule a refresh immediately with no delay
+            self._next = self._loop.create_task(self._schedule_refresh(0))
+            self._current = await self._next
+            return True
+        except Exception as e:
+            # if anything else goes wrong, log the error and return false
+            logger.exception("Error occurred during force refresh attempt", exc_info=e)
+            return False
+
+    def force_refresh(self, timeout: Optional[int] = None) -> bool:
```
kurtisvg, line 376: nit: This function should probably have a docstring

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,95 @@
...
+    :type loop: asyncio.AbstractEventLoop
+    :param: loop:
+        The event loop to use. If not provided, the default event loop will be used.
+
+
+    """
+
+    def __init__(
+        self,
+        burst_size: int = 1,
```
kurtisvg, line 44: Do we ever need burst_size to be greater than one? It think we should never need to acquire more than one token at a time

google/cloud/sql/connector/rate_limiter.py
```diff
@@ -0,0 +1,95 @@
...
+        self._last_token_update = time.time()
+
+    def _add_tokens(self) -> None:
+        """
+        Calculates how much time has passed since the last leak and removes the
+        appropriate amount of events from the queue.
+        Leaking is done lazily, meaning that if there is a large time gap between
+        leaks, the next set of calls might be a burst if burst_size > 1
+        """
+        time_elapsed = time.time() - self._last_token_update
```
kurtisvg, line 61: We should use `time.monotonic`, maybe even asncyio.time?

google/cloud/sql/connector/rate_limiter.py
kurtisvg: I think maybe we should chat about this, because I think we should be able to eliminate some of the complexity here. Specifically it feels like the logic should be something like: 

```
async with lock
    calculate updated token count
    if tokens <= 0:
        wait until next token is ready
        calculate updated token count
    wait until time the next token is ready
    update token count
    tokens -= 1
        // success!
```

kurtisvg (changes requested):

google/cloud/sql/connector/instance_connection_manager.py
kurtisvg: I think this would still be a good idea

kurtisvg (approved):

google/cloud/sql/connector/instance_connection_manager.py
shubha-rajan: added a docstring

PR #147, opened 2021-08-13
chore: add region tag for IAM auth connection code

shubha-rajan:


kurtisvg (approved):
We should include these in the v2 samples

PR #178, opened 2021-10-05
chore: updating to 0.4.2-dev

jackwotherspoon:


google/cloud/sql/connector/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.4.1"
+__version__ = "0.4.1-dev"
```
shubha-rajan, line 15: shouldn't this be "0.4.2-dev"?

shubha-rajan (approved):

PR #196, opened 2021-11-04
docs: update README with connection pooling

### LINKED ISSUE #195: Update README to recommend sqlalchemy implementation by default.
Update README to mention and recommend using sqlalchemy to manage connection pooling. It currently doesn't explain why in the sqlalchemy code sample.

jackwotherspoon:
Closes #195

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,9 @@ for row in result:
     print(row)
 ```
 
+#### Using a Connection Pool (RECOMMENDED)
+The connector itself creates connection objects but does not manage database connection pooling. For this reason, it is recommended to use the connector alongside a library that can create connection pools, such as [SQLAlchemy](https://www.sqlalchemy.org/). This will allow for connections to remain open and be reused, reducing connection overhead and the number of connections needed.
+
 To use this connector with SQLAlchemy, use the `creator` argument for `sqlalchemy.create_engine`:
 ```
 def getconn() -> pymysql.connections.Connection:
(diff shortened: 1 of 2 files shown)


enocom (approved):

README.md
```diff
@@ -94,6 +97,7 @@ engine = sqlalchemy.create_engine(
     creator=getconn,
 )
 ```
+**Note**: For more examples of using SQLAlchemy to manage connection pooling with the connector, please see [Cloud SQL SQLAlchemy Samples](https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/cloud-sql).
```
kurtisvg, line 100: Let's use this page instead https://cloud.google.com/sql/docs/mysql/manage-connections#python

README.md
```diff
@@ -77,6 +77,9 @@ for row in result:
     print(row)
 ```
 
+#### Using a Connection Pool (RECOMMENDED)
```
kurtisvg, line 80: I'd like to see this part first. Ideally the person coming into the "How to use this connector" would receive a recommendation to use a connection pool within the first 2-3 sentences. 

I think this means we'd probably also remove the example without the pool, and show an example using the pool instead.

kurtisvg (changes requested):

README.md
jackwotherspoon: How about this page as it shows the postgres version of the code sample? The link you recommended and the one I originally had don't show SQLAlchemy with the connector.
https://cloud.google.com/sql/docs/postgres/connect-connectors#python_1

README.md
```diff
@@ -45,42 +45,20 @@ pip install git+https://github.com/GoogleCloudPlatform/cloud-sql-python-connecto
...
-for row in result:
-    print(row)
-```
+In the connector's `connect` method below, input your connection string as the first positional argument and the name of the database driver for the second positional argument. Insert the rest of your connection keyword arguments like user, password and database. You can also set the optional `timeout` or `ip_type` keyword arguments.
 
 To use this connector with SQLAlchemy, use the `creator` argument for `sqlalchemy.create_engine`:
 ```
-def getconn() -> pymysql.connections.Connection:
-    conn: pymysql.connections.Connection = connector.connect(
+def getconn():
```
kurtisvg, line 60: Why did we drop the typing?

kurtisvg (approved):

README.md
jackwotherspoon: I can add it back in if you want? I just thought it made the sample look a little cluttered. I'm open to adding it back in.

kurtisvg: I like typing. It's usually valuable if you're using typing, and easy to ignore if you aren't.

PR #210, opened 2021-12-09
feat: expose Connector object to allow multiple connection configurations.

jackwotherspoon:


google/cloud/sql/connector/connector.py
```diff
@@ -25,41 +25,125 @@
...
-    driver: str,
-    ip_types: IPTypes = IPTypes.PUBLIC,
-    enable_iam_auth: bool = False,
-    **kwargs: Any
-) -> Any:
-    """Prepares and returns a database connection object and starts a
-    background thread to refresh the certificates and metadata.
+    def __init__(
+        self,
+        ip_types: IPTypes = IPTypes.PUBLIC,
```
kurtisvg, line 51: nit: I think technically this should be "ip_type"?

google/cloud/sql/connector/connector.py
```diff
@@ -25,41 +25,125 @@
...
+        enable_iam_auth: bool = False,
+        timeout: int = 30,
+    ) -> None:
+        # This thread is used for background processing
+        self._thread: Optional[Thread] = None
+        self._loop: Optional[asyncio.AbstractEventLoop] = None
+        self._keys: Optional[concurrent.futures.Future] = None
```
kurtisvg, line 58: Setting these to None was necessary before because they were global state. I think we can initialize by default since they are inside the object now.

google/cloud/sql/connector/connector.py
```diff
@@ -25,41 +25,125 @@
...
+            timeout = self._timeout
+        try:
+            return icm.connect(driver, ip_types, timeout, **kwargs)
+        except Exception as e:
+            # with any other exception, we attempt a force refresh, then throw the error
+            icm.force_refresh()
+            raise (e)
+
+
+_default_connector: Optional[Connector] = None
```
kurtisvg, line 140: nit: Let's try to get global state to the top of the file so it's harder to miss

google/cloud/sql/connector/connector.py
```diff
@@ -25,41 +25,125 @@
...
+
+
+def connect(instance_connection_string: str, driver: str, **kwargs: Any) -> Any:
+    """Creates default Connector object and returns a database connection object
```
kurtisvg, line 144:
```suggestion
    """Uses a  Connector object with default settings and returns a database connection object
```

kurtisvg (changes requested):
There are some tests that we should update to use a Connector object. That can be done in a separate PR w/ mocking if you want, just want to make sure it's on your radar.

google/cloud/sql/connector/connector.py
```diff
@@ -23,43 +23,118 @@
...
+            icm = InstanceConnectionManager(
+                instance_connection_string,
+                driver,
+                self._keys,
+                self._loop,
+                enable_iam_auth,
+            )
+            self._instances[instance_connection_string] = icm
+
+        ip_type = kwargs.pop("ip_types", self._ip_type)
```
jackwotherspoon, line 119: @kurtisvg What would you recommend here? Since the current version of the Python connector takes in `ip_types` as the parameter. Should I add support for both by adding a `kwargs.pop("ip_type")` check as well? Then update the README etc to use `ip_type` as the golden standard. Leaving in `ip_types` so we are still backwards compatible and doesn't break anything for current users.

kurtisvg: That's fine. You may want to add a log warning that ip_types is deprecated and we may remove it in a future release.

kurtisvg (approved):

PR #224, opened 2021-12-16
chore: add versioning of dependencies in noxfile

jackwotherspoon:


noxfile.py
```diff
@@ -17,24 +17,30 @@
...
-        "types-pkg-resources", "types-PyMySQL", "twine")
+        "flake8==4.0.1",
+        "flake8-annotations==2.7.0",
+        "black==21.12b0",
+        "mypy==0.910",
+        "sqlalchemy-stubs==0.4",
+        "types-pkg-resources==0.1.3",
+        "types-PyMySQL==1.0.6",
+        "twine==3.7.1",
+    )
```
kurtisvg, line 43: Do you have a plan for how we want to update these? presumably renovate won't bump them. Should we put them in a different spot where it can?

kurtisvg (changes requested):

kurtisvg (approved):

PR #225, opened 2021-12-17
update classifiers

anukaal:
added support for python 3.10


jackwotherspoon:
Hi @anukaal Thanks so much for opening up this PR! It is definitely important that the python connector support python 3.10. This code change was a great start and very helpful. I've included it in #227 as well as added a few additional changes to fully support 3.10 such as tests and updating some deprecated asyncio loop params. Closing this PR as it is now a duplicate.

Feel free to contribute more updates in the future! We always love contributions from individual contributors :) Thanks!

PR #226, opened 2021-12-21
feat: add arg for specifying credentials

jackwotherspoon:


google/cloud/sql/connector/connector.py
```diff
@@ -43,16 +43,23 @@ class Connector:
...
+        or google.auth.credentials.Credentials object.
+        If not specified, Application Default Credentials are used.
     """
 
     def __init__(
         self,
         ip_types: IPTypes = IPTypes.PUBLIC,
         enable_iam_auth: bool = False,
         timeout: int = 30,
+        service_account_creds: Union[str, Credentials, None] = None,
```
kurtisvg, line 62: This should just be "credentials". It's not always a service account

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -343,17 +373,31 @@ async def _get_instance_data(self) -> InstanceMetadata:
...
+            "https://www.googleapis.com/auth/cloud-platform",
+        ]
+        # if Credentials object is passed in, use for authentication
+        if isinstance(service_account_creds, Credentials):
+            credentials = with_scopes_if_required(service_account_creds, scopes=scopes)
+        # if string is passed in, load credentials from file
+        elif isinstance(service_account_creds, str):
+            credentials, project = google.auth.load_credentials_from_file(
+                filename=service_account_creds, scopes=scopes
+            )
```
kurtisvg, line 397: Hmm this seems pretty simple. Would we be better off just having the user do this and only taking in a `credentials` type?

kurtisvg (changes requested):

google/cloud/sql/connector/instance_connection_manager.py
jackwotherspoon: That works for me and makes it a little more straightforward to understand for users. I was just trying to mimic the go connector that allows credentials from JSON. I'll make it take in credentials, otherwise use default creds.

kurtisvg: That makes sense. I think for python, it would make more sense to either provide a helper function that loads the credential file with the correct scopes or just rescope the credentials after we get them.

kurtisvg (approved):

PR #227, opened 2021-12-22
feat: add support for python 3.10

jackwotherspoon:


google/cloud/sql/connector/rate_limiter.py
```diff
@@ -46,7 +47,12 @@ def __init__(
         self.rate = rate
         self.max_capacity = max_capacity
         self._loop = loop or asyncio.get_event_loop()
-        self._lock = asyncio.Lock(loop=self._loop)
+        self._lock = asyncio.Lock()
+        # set based on python version
+        if sys.version_info[:3] >= (3, 8):
+            self._lock = asyncio.Lock()
+        else:
+            self._lock = asyncio.Lock(loop=self._loop)
```
kurtisvg, line 55: Hmm this seems problematic. Do we need the loop or not?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -258,7 +259,11 @@ def __init__(
 
         async def _set_instance_data() -> None:
             logger.debug("Updating instance data")
-            self._refresh_in_progress = asyncio.locks.Event(loop=self._loop)
+            # set based on python version
+            if sys.version_info[:3] >= (3, 8):
+                self._refresh_in_progress = asyncio.locks.Event()
+            else:
+                self._refresh_in_progress = asyncio.locks.Event(loop=self._loop)
```
kurtisvg, line 266: Same here. Why do we want this to behave different on two different python versions?

kurtisvg (changes requested):

google/cloud/sql/connector/rate_limiter.py
jackwotherspoon: @kurtisvg I'll look into it a bit deeper. Basically the loop param for some asyncio functions were deprecated in Python 3.8 and removed in 3.10 so my thinking was having Python 3.8+ remove it and earlier versions keep it. Let me know what you think is best. I can remove it for all and see if tests pass.

kurtisvg: It just depends on why the loop is being removed. if we need it for old versions of python that's one thing, but it the behavior is the same it's better to just leave it so it's easier to debug. So we should determine:

1. Why they are removing the value?
2. Do we _want_ it to execute in that loop? (or doesn't it matter?)

From there we can decide if it's better to have split functionality or singular.

jackwotherspoon: Based on reading https://stackoverflow.com/questions/60312374/what-are-all-these-deprecated-loop-parameters-in-asyncio and a few other discussions, it seems that Python 3.6 fixed an issue that makes the loop parameter no longer needed. Since we only support Python versions > 3.6 we should be fine just removing the loop param and not needing the split functionality.

The methods that used the `loop` param now just implicitly get the current running loop.

kurtisvg (approved):

PR #241, opened 2022-01-12
fix: update locks to be thread-safe

jackwotherspoon:
Current locking code `asyncio.Lock()` is not thread-safe and needs to be replaced with `threading.Lock()`


kurtisvg:
The strategy up until now has been to use a single background thread for the event loop, which means we only need to worry about thread safety when interacting with the background thread. Do you have a reason to think we need a better locking here?

PR #252, opened 2022-01-26
fix: make asyncio.Lock() run in background thread

jackwotherspoon:
@shubha-rajan Would love to have you take a glance at this when you have a free second :)

### DIFF
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -279,7 +279,6 @@ class InstanceConnectionManager:
             """Initialize InstanceConnectionManager's variables that require the
             event loop running in background thread.
             """
-            logger.debug("Updating instance data")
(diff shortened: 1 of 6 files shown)


google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -274,17 +275,19 @@ def __init__(
...
-        self._refresh_rate_limiter = AsyncRateLimiter(
-            max_capacity=2, rate=1 / 30, loop=self._loop
-        )
-
-        async def _set_instance_data() -> None:
+        async def _async_init() -> None:
+            """Initialize InstanceConnectionManager's variables that require the
+            event loop running in background thread.
+            """
             logger.debug("Updating instance data")
```
kurtisvg, line 282: nit: I think we can probably remove this log line?

kurtisvg (approved):

PR #273, opened 2022-02-10
fix: remove enable_iam_auth from downstream kwargs and catch error

### LINKED ISSUE #272: Conflicting enable_iam_auth argument causes several errors

jackwotherspoon:
Closes #272

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -108,11 +108,18 @@ class Connector:
         # Use the InstanceConnectionManager to establish an SSL Connection.
         #
         # Return a DBAPI connection
-
+        enable_iam_auth = kwargs.pop("enable_iam_auth", self._enable_iam_auth)
         if instance_connection_string in self._instances:
             icm = self._instances[instance_connection_string]
+            if enable_iam_auth != icm._enable_iam_auth:
+                raise ValueError(
+                    "Connect method got `enable_iam_auth={}`, but previously got "
+                    "enable_iam_auth={}`. If you require both for your use-case, "
+                    "please use a new connector.Connector object.".format(
+                        enable_iam_auth, icm._enable_iam_auth
+                    )
+                )
         else:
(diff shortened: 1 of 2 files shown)


tests/unit/test_connector.py
```diff
@@ -59,6 +59,33 @@ async def timeout_stub(*args: Any, **kwargs: Any) -> None:
...
+        connector._default_connector = mock_connector
+        # try to connect using enable_iam_auth=True
+        with patch.dict(mock_connector._instances, mock_instances):
+            pytest.raises(
+                ValueError,
+                connector.connect,
+                connect_string,
+                "pymysql",
+                enable_iam_auth=True,
+            )
```
kurtisvg, line 86: This test would be better if you tested it a little higher up, as it would more concretely check the logic of how a user would see it. Can you initialize a connector and call connect twice on it?

google/cloud/sql/connector/connector.py
```diff
@@ -108,11 +108,18 @@ def connect(
         # Use the InstanceConnectionManager to establish an SSL Connection.
         #
         # Return a DBAPI connection
-
+        enable_iam_auth = kwargs.pop("enable_iam_auth", self._enable_iam_auth)
         if instance_connection_string in self._instances:
             icm = self._instances[instance_connection_string]
+            if enable_iam_auth != icm._enable_iam_auth:
+                raise ValueError(
+                    "Connect method got `enable_iam_auth={}`, but previously got "
```
kurtisvg, line 116:
```suggestion
                    "connect() called with `enable_iam_auth={}`, but previously used "
```

google/cloud/sql/connector/connector.py
```diff
@@ -108,11 +108,18 @@ def connect(
...
+                    "enable_iam_auth={}`. If you require both for your use-case, "
```
kurtisvg, line 117:
```suggestion
                    "enable_iam_auth={}`. If you require both for your use case, "
```

kurtisvg (changes requested):

kurtisvg (approved):

PR #280, opened 2022-02-24
feat: add asynchronous interface to Connector

jackwotherspoon:
Add the async interface to the Connector.

Will add the logic to choose the proper event loop when I add support for the first async driver.


google/cloud/sql/connector/connector.py
```diff
@@ -137,17 +166,21 @@ def connect(
...
+                icm.connect(driver, ip_type, **kwargs)
+            )
+            await asyncio.wait_for(connection_task, timeout)
+            return await connection_task
+        except asyncio.TimeoutError:
+            raise TimeoutError(f"Connection timed out after {timeout}s")
         except Exception as e:
             # with any other exception, we attempt a force refresh, then throw the error
-            icm.force_refresh()
+            await self._loop.create_task(icm.force_refresh())
```
jackwotherspoon, line 183: @kurtisvg should this just be `await icm.force_refresh()`? There is no benefit of creating a task to await it immediately right? Just double checking.

google/cloud/sql/connector/connector.py
```diff
@@ -137,17 +166,21 @@ def connect(
...
+        if "connect_timeout" in kwargs:
             timeout = kwargs["connect_timeout"]
-        else:
-            timeout = self._timeout
+
         try:
-            return icm.connect(driver, ip_type, timeout, **kwargs)
+            connection_task = self._loop.create_task(
+                icm.connect(driver, ip_type, **kwargs)
+            )
```
kurtisvg, line 176: I don't think this needs to be a task anymore since we are already in async land by this point. You can await for directly on the coro generated by calling the function.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -266,19 +266,12 @@ def __init__(
...
-            self._next = self._loop.create_task(self._schedule_refresh())
-
-        init_future = asyncio.run_coroutine_threadsafe(_async_init(), self._loop)
-        init_future.result()
+        self._refresh_rate_limiter = AsyncRateLimiter(
+            max_capacity=2, rate=1 / 30, loop=self._loop
+        )
+        self._refresh_in_progress = asyncio.locks.Event()
+        self._current = self._loop.create_task(self._get_instance_data())
+        self._next = self._loop.create_task(self._schedule_refresh())
```
kurtisvg, line 274: I think we already talked about this, but I think schedule_refresh should return a task directly.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -365,26 +366,14 @@ async def _force_refresh(self) -> bool:
             self._next.cancel()
             # schedule a refresh immediately with no delay
             self._next = self._loop.create_task(self._schedule_refresh(0))
+            await asyncio.wait_for(self._next, timeout)
```
kurtisvg, line 369: I think we already talked about this too, but this should be a instant sync operation. It need to trigger the next refresh, but tot await on the results.

kurtisvg (changes requested):
I think we may want to take a bottom out approach here and make some of the lower level changes we discussed first, then come back and add the connect interface.

kurtisvg (approved):

Issue #103, opened 2021-05-14
Clean up perform_refresh and schedule_refresh methods


kurtisvg:
These methods are a bit confusing - see if we can make them a bit cleaner now that we aren't using locking on the backend. 
- Update `_perform_refresh` to return an `InstanceMetadata` object instead of a task to eliminate use of nested tasks
- Move static methods to their own files

kurtisvg:
Discussing this today, we decided we need to refactor the following: `_perform_refresh`, `_schedule_refresh`, `_get_instancedata`.

Refactor into 2 functions: 

- _schedule_refresh ->
    -  create the task and returns that task
    - Task:
        - Sleep for duration
        - Run perform refresh
        - If perform refresh is bad
            - Schedule new run immediately
            - If the old value is also bad, replace old value
        - Schedule the next refresh that happens 
- _perform_refresh -> perform the operation

PR #294, opened 2022-03-09
chore: refactor refresh methods

jackwotherspoon:
Refactor refresh methods to match those of the [Cloud SQL Go Connector](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector).

Separate PR will make refresh methods into static methods within `refresh_utils.py`

### DIFF
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -116,6 +116,15 @@ class CredentialsTypeError(Exception):
         super(CredentialsTypeError, self).__init__(self, *args)
 
 
+class ExpiredInstanceMetadata(Exception):
+    """
+    Raised when InstanceMetadata is expired.
+    """
+
+    def __init__(self, *args: Any) -> None:
+        super(CredentialsTypeError, self).__init__(self, *args)
+
+
 class InstanceMetadata:
     ip_addrs: Dict[str, Any]
     context: ssl.SSLContext
@@ -274,8 +283,8 @@ class InstanceConnectionManager:
                 max_capacity=2, rate=1 / 30, loop=self._loop
             )
             self._refresh_in_progress = asyncio.locks.Event()
-            self._current = self._loop.create_task(self._perform_refresh())
-            self._next = self._schedule_refresh()
+            self._current = self._schedule_refresh(0)
+            self._next = self._current
 
         init_future = asyncio.run_coroutine_threadsafe(_async_init(), self._loop)
         init_future.result()
@@ -323,7 +332,13 @@ class InstanceConnectionManager:
             logger.exception("Error occurred during force refresh attempt", exc_info=e)
             return False
 
-    async def seconds_until_refresh(self) -> int:
+    async def _seconds_until_refresh(self) -> int:
+        """
+        Helper function to get time in seconds before performing next refresh.
+
+        :rtype: int
+        :returns: Time in seconds to wait before performing next refresh.
+        """
         expiration = (await self._current).expiration
 
         if self._enable_iam_auth:
@@ -415,27 +430,30 @@ class InstanceConnectionManager:
             self._enable_iam_auth,
         )
 
-    def _schedule_refresh(self, delay: Optional[int] = None) -> asyncio.Task:
+    def _schedule_refresh(self, delay: int) -> asyncio.Task:
         """
         Schedule task to sleep and then perform refresh to get InstanceMetadata.
+
+        :type delay: int
+        :param delay
+            Time in seconds to sleep before running _perform_refresh.
 
         :rtype: asyncio.Task
         :returns: A Task representing the scheduled _perform_refresh.
         """
 
         async def _refresh_task(
-            self: InstanceConnectionManager, delay: Optional[int]
+            self: InstanceConnectionManager, delay: int
         ) -> InstanceMetadata:
             """
             A coroutine that sleeps for the specified amount of time before
             running _perform_refresh.
             """
-            if delay is None:
-                delay = await self.seconds_until_refresh()
-            refresh_task = self._loop.create_task(self._perform_refresh())
+            refresh_task = None
             try:
                 logger.debug("Entering sleep")
                 await asyncio.sleep(delay)
+                refresh_task = self._loop.create_task(self._perform_refresh())
                 await refresh_task
             except asyncio.CancelledError as e:
                 logger.debug("Schedule refresh task cancelled.")
@@ -448,24 +466,24 @@ class InstanceConnectionManager:
                     exc_info=e,
                 )
                 # check if current metadata is valid, don't want to replace valid metadata with invalid one
-                instance_data = None
                 try:
                     instance_data = await self._current
+                    if instance_data.expiration < datetime.datetime.now():
+                        raise ExpiredInstanceMetadata(
+                            "Instance metadata is expired, performing new refresh."
+                        )
                 except Exception:
                     # Current result is invalid, no-op
                     logger.debug("Current instance data is invalid.")
-                # if current metadata is invalid, replace with more recent invalid refresh and schedule another refresh immediately
-                if (
-                    instance_data is None
-                    or instance_data.expiration < datetime.datetime.now()
-                ):
                     self._current = refresh_task
-                    self._next = self._schedule_refresh(0)
+                # schedule new refresh attempt immediately
+                self._next = self._schedule_refresh(0)
             # if valid refresh, replace current with valid metadata and schedule next refresh
             else:
                 self._current = refresh_task
                 # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
-                self._next = self._schedule_refresh()
+                delay = await self._seconds_until_refresh()
+                self._next = self._schedule_refresh(delay)
 
             return await refresh_task
 
--- a/google/cloud/sql/connector/instance_connection_manager.py
+++ b/google/cloud/sql/connector/instance_connection_manager.py
@@ -327,10 +327,10 @@ class InstanceConnectionManager:
             and a string representing a PEM-encoded certificate authority.
         """
         self._refresh_in_progress.set()
-        await self._refresh_rate_limiter.acquire()
         logger.debug("Entered _perform_refresh")
 
         try:
+            await self._refresh_rate_limiter.acquire()
             priv_key, pub_key = await self._keys
 
             logger.debug("Creating context")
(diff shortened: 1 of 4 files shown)


tests/unit/test_instance_connection_manager.py
```diff
@@ -28,7 +29,38 @@
...
+
+async def _get_metadata_success(*args: Any, **kwargs: Any) -> MockMetadata:
+    return MockMetadata(datetime.datetime.now() + datetime.timedelta(minutes=10))
+
+
+async def _get_metadata_error(*args: Any, **kwargs: Any) -> None:
+    raise BadRefresh("something went wrong...")
+
+
+async def refresh(icm: InstanceConnectionManager, wait: bool = False) -> asyncio.Task:
```
jackwotherspoon, line 49: Temporary solution: Will be removed with async interface PR when tests will run with same event loop as ICM and not need `run_coroutine_threadsafe`

tests/unit/test_instance_connection_manager.py
```diff
@@ -28,7 +29,38 @@
...
+    """
+    Helper function to run schedule_refresh within background thread's event_loop
+    """
+    new_task = icm._schedule_refresh(0)
+
+    # since we can't await in test itself we set flag to await in background thread for result
+    if wait:
+        await new_task
+
+    return new_task
```
jackwotherspoon, line 59: Temporary solution: Will be removed with async interface PR when tests will run with same event loop as ICM and not need `run_coroutine_threadsafe`

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -406,66 +344,134 @@ async def seconds_until_refresh(self) -> int:
...
+                ):
+                    self._current = refresh_task
+                    self._next = self._schedule_refresh(0)
+            # if valid refresh, replace current with valid metadata and schedule next refresh
+            else:
+                self._current = refresh_task
+                # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
+                self._next = self._schedule_refresh()
+
+            return await refresh_task
```
jackwotherspoon, line 470: Is it okay to set `refresh_task` to current and next within here? Wonder if there is a way to not need the intermediary task of `refresh_task` and only deal with `scheduled_task`.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -406,66 +344,134 @@ async def seconds_until_refresh(self) -> int:
...
-        return await self._perform_refresh()
+        async def _refresh_task(
+            self: InstanceConnectionManager, delay: Optional[int]
+        ) -> InstanceMetadata:
+            """
+            A coroutine that sleeps for the specified amount of time before
+            running _perform_refresh.
+            """
+            if delay is None:
+                delay = await self.seconds_until_refresh()
```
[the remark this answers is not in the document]
kurtisvg: Actually, I think would be better to move this calculation to happen as part of the perform refresh and make `delay` a required variable. There's really no reason to have `seconds_until_refresh` be a public interface for the ICM - we only use it in the one place, and I think it's more clear to explicitly pass in the value.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -406,66 +344,134 @@ async def seconds_until_refresh(self) -> int:
...
+                    instance_data is None
+                    or instance_data.expiration < datetime.datetime.now()
+                ):
+                    self._current = refresh_task
+                    self._next = self._schedule_refresh(0)
+            # if valid refresh, replace current with valid metadata and schedule next refresh
+            else:
+                self._current = refresh_task
+                # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
+                self._next = self._schedule_refresh()
```
kurtisvg, line 468: I think we want this logic to be:

```python
if self._current is invalid: # expired
     # self._current is not useful anyway, so let's return the latest error
     self._current = refresh_task
# always schedule a new refresh if the last one was bad 
self._next = self._schedule_refresh(0)
```

kurtisvg (changes requested):

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -274,66 +274,12 @@ async def _async_init() -> None:
                 max_capacity=2, rate=1 / 30, loop=self._loop
             )
             self._refresh_in_progress = asyncio.locks.Event()
-            self._current = self._loop.create_task(self._get_instance_data())
-            self._next = self._loop.create_task(self._schedule_refresh())
+            self._current = self._loop.create_task(self._perform_refresh())
+            self._next = self._schedule_refresh()
```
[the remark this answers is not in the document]
jackwotherspoon: how come we want to set `self._next = self._current` here? Are we just trying to say let `schedule_refresh` deal with setting `self._next`? If that is the case we probably don't even need next to be set here.

google/cloud/sql/connector/instance_connection_manager.py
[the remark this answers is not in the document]
jackwotherspoon: If we were to move `delay` into `perform_refresh` then how would `schedule_refresh` know how long to sleep for?

google/cloud/sql/connector/instance_connection_manager.py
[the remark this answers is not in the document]
kurtisvg: I think you want to set `_next` here so that you can never have `self._next` be `None`.

google/cloud/sql/connector/instance_connection_manager.py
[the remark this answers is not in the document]
kurtisvg: `schedule_refresh` takes delay as an argument (as opposed to optional). 

In `perform_refresh` where we call `schedule_refresh`, we pass in delay (as opposed to using the "default"). `perform_refresh` can do the calculation for delay in the function, as opposed to relying on `seconds_until_refresh`.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -356,36 +311,34 @@ def _auth_init(self, credentials: Optional[Credentials]) -> None:
...
             # schedule a refresh immediately with no delay
-            self._next = self._loop.create_task(self._schedule_refresh(0))
-            self._current = await self._next
+            self._next = self._schedule_refresh(0)
+            self._current = self._next
             return True
         except Exception as e:
             # if anything else goes wrong, log the error and return false
             logger.exception("Error occurred during force refresh attempt", exc_info=e)
             return False
```
kurtisvg, line 333: This section is feeling a bit wonky. Where do we expect an `Exception` to be thrown? Is it just `cancel()`? If so I think we should move L327-L329 out of the try/catch. I don't think we want to return false if we actually scheduled a refresh (but I don't think we can get an error there anymore)

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -356,36 +311,34 @@ def _auth_init(self, credentials: Optional[Credentials]) -> None:
 
         self._credentials = credentials
 
-    async def _force_refresh(self) -> bool:
+    def force_refresh(self) -> bool:
+        """
+        Forces a new refresh attempt and returns a boolean value that indicates
+        whether the attempt was successful.
+        """
+        # if a new refresh is already in progress, then block on the result
```
kurtisvg, line 319:
```suggestion
        # if a new refresh is already in progress, then block new connections on the next result
```

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -406,66 +359,138 @@ async def seconds_until_refresh(self) -> int:
...
+        return InstanceMetadata(
+            ephemeral_cert,
+            metadata["ip_addresses"],
+            priv_key,
+            metadata["server_ca_cert"],
+            expiration,
+            self._enable_iam_auth,
+        )
+
+    def _schedule_refresh(self, delay: int) -> asyncio.Task:
```
kurtisvg, line 433: nit: should this be `asyncio.Task[InstanceMetadata]`?

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -406,66 +359,138 @@ async def seconds_until_refresh(self) -> int:
...
+                    # Current result is invalid, no-op
+                    logger.debug("Current instance data is invalid.")
+                    self._current = refresh_task
+                # schedule new refresh attempt immediately
+                self._next = self._schedule_refresh(0)
+            # if valid refresh, replace current with valid metadata and schedule next refresh
+            else:
+                self._current = refresh_task
+                # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
+                delay = await self._seconds_until_refresh()
```
kurtisvg, line 486: Can we make `_seconds_until_refresh` either something on InstanceMetadata or a standalone function? I don't think it needs access to `self` here.

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -406,66 +359,138 @@ async def seconds_until_refresh(self) -> int:
...
+                    "Scheduling another refresh attempt immediately",
+                    exc_info=e,
+                )
+                # check if current metadata is valid, don't want to replace valid metadata with invalid one
+                try:
+                    instance_data = await self._current
+                    if instance_data.expiration < datetime.datetime.now():
+                        raise ExpiredInstanceMetadata(
+                            "Instance metadata is expired, performing new refresh."
+                        )
```
kurtisvg, line 475: I feel like this bit is making the rest hard to read. What do you think about an `is_valid` function to help here:

```python
def is_valid(task: asyncio.Task[InstanceMetadata]) -> bool:
     try: 
          md = await task
         # only valid if now is before the cert expires
          if datetime.datetime.now() < instance_data.expiration:
              return True
      except Exception:
           # surpress any errors from task
      return False
```

Should help simplify this try/catch nest quite a bit

google/cloud/sql/connector/instance_connection_manager.py
kurtisvg: I'm also not sure we need another context switch here - we should have the result of `instance_data` from the await of `refresh_task` earlier, which should be what we need to calculate the delay

google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -406,66 +359,138 @@ async def seconds_until_refresh(self) -> int:
...
+                # schedule new refresh attempt immediately
+                self._next = self._schedule_refresh(0)
+            # if valid refresh, replace current with valid metadata and schedule next refresh
+            else:
+                self._current = refresh_task
+                # Ephemeral certificate expires in 1 hour, so we schedule a refresh to happen in 55 minutes.
+                delay = await self._seconds_until_refresh()
+                self._next = self._schedule_refresh(delay)
+
+            return await refresh_task
```
kurtisvg, line 489: We're already awaiting this on L458, we should save that value and return it here (rather than an extra context switch)

kurtisvg (changes requested):

kurtisvg (approved):

jackwotherspoon:
Closes #103

PR #304, opened 2022-03-22
chore: refactor connect into connect_info

### LINKED ISSUE #290: Refactor "connect" into "connect_info"
Update "ICM.connect" to "ICM.connect_info"

This should also move the "connect_with_XXXX" functions up to the connector level as well.

jackwotherspoon:
Closes #290

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -27,7 +27,7 @@ from google.cloud.sql.connector.instance_connection_manager import (
 from google.cloud.sql.connector.utils import generate_keys, clean_input
 from google.auth.credentials import Credentials
 from threading import Thread
-from typing import Any, Dict, Optional, TYPE_CHECKING
+from typing import Any, Dict, Optional, TYPE_CHECKING, Callable
 from functools import partial
 
 logger = logging.getLogger(name=__name__)
@@ -186,12 +186,8 @@ class Connector:
 
         # attempt to make connection to Cloud SQL instance for given timeout
         try:
-            instance_data, ip_address = await asyncio.wait_for(icm.connect_info(ip_type), timeout)
-            connect_partial = partial(
-                connector, ip_address, instance_data.context, **kwargs
-            )
             return await asyncio.wait_for(
-                self._loop.run_in_executor(None, connect_partial), timeout
+                get_connection(icm, connector, ip_type, self._loop, **kwargs), timeout
             )
         except asyncio.TimeoutError:
             raise TimeoutError(f"Connection timed out after {timeout}s")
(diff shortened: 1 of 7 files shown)


google/cloud/sql/connector/instance_connection_manager.py
```diff
@@ -438,171 +429,31 @@ async def _refresh_task(
...
-                raise PlatformNotSupportedError(
-                    "Active Directory authentication is currently only supported on Windows."
-                )
-
-        user = kwargs.pop("user")
-        passwd = kwargs.pop("password")
-        return pytds.connect(
-            ip_address, database=db, user=user, password=passwd, sock=sock, **kwargs
-        )
+        return (instance_data, ip_address)
```
kurtisvg, line 456: nit: You don't need to explicitly bundle this in a tuple, python does it automatically with just `return a, b`

google/cloud/sql/connector/utils.py
```diff
@@ -77,3 +80,27 @@ def write_to_file(
         priv_out.write(priv_key)
 
     return (ca_filename, cert_filename, key_filename)
+
+
+def clean_input(default_ip_type: Any, default_timeout: int, **kwargs: Any) -> Any:
```
kurtisvg, line 85: We're only using this in the 1 place - is it really worth separating out into it's own function?

google/cloud/sql/connector/connector.py
```diff
@@ -229,3 +243,169 @@ def connect(instance_connection_string: str, driver: str, **kwargs: Any) -> Any:
...
+        else:
+            raise PlatformNotSupportedError(
+                "Active Directory authentication is currently only supported on Windows."
+            )
+
+    user = kwargs.pop("user")
+    passwd = kwargs.pop("password")
+    return pytds.connect(
+        ip_address, database=db, user=user, password=passwd, sock=sock, **kwargs
+    )
```
kurtisvg, line 411: nit: these 3 functions might be easier to find if we move the into their own files

google/cloud/sql/connector/connector.py
```diff
@@ -229,3 +243,169 @@ def connect(instance_connection_string: str, driver: str, **kwargs: Any) -> Any:
...
     return _default_connector.connect(instance_connection_string, driver, **kwargs)
+
+
+async def get_connection(
+    icm: InstanceConnectionManager,
+    connect_function: Callable,
+    ip_type: IPTypes,
+    loop: asyncio.AbstractEventLoop,
+    **kwargs: Any,
+) -> Any:
```
kurtisvg, line 254: I don't really understand this function's signature (or why we need it). It's also very short (minus the docstring). Are we sure we shouldn't inline here?

kurtisvg (changes requested):

google/cloud/sql/connector/utils.py
```diff
@@ -19,6 +19,9 @@
 from cryptography.hazmat.primitives.asymmetric import rsa
 
 from typing import Tuple
+import logging
+
+logger = logging.getLogger(name=__name__)
```
kurtisvg, line 24: Are we using these here?

kurtisvg (changes requested):

google/cloud/sql/connector/connector.py
```diff
@@ -170,10 +186,31 @@ async def connect_async(
...
+        kwargs.pop("ssl", None)
+        kwargs.pop("port", None)
+
+        # helper function to wrap in timeout
+        async def get_connection(
+            self: Connector,
+            icm: InstanceConnectionManager,
+            connect_function: Callable,
+            ip_type: IPTypes,
+            **kwargs: Any,
```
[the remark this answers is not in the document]
jackwotherspoon: Removed need for these arguments.

google/cloud/sql/connector/connector.py
```diff
@@ -171,11 +187,24 @@ async def connect_async(
...
 
+        # Host and ssl options come from the certificates and metadata, so we don't
+        # want the user to specify them.
+        kwargs.pop("host", None)
+        kwargs.pop("ssl", None)
+        kwargs.pop("port", None)
+
+        # helper function to wrap in timeout
+        async def get_connection() -> Any:
+
```
kurtisvg, line 198: nit: 
```suggestion
```

kurtisvg (approved):

PR #315, opened 2022-04-04
chore: update InstanceConnectionManager to Instance

jackwotherspoon:
Renaming `InstanceConnectionManager` class to `Instance` as it no longer manages connections, that has been moved up a level to the connector.

### DIFF
--- a/google/cloud/sql/connector/__init__.py
+++ b/google/cloud/sql/connector/__init__.py
@@ -17,7 +17,7 @@ limitations under the License.
 from typing import List
 
 from .connector import connect, Connector
-from .instance_connection_manager import IPTypes
+from .instance import IPTypes
 
 
 __ALL__ = [connect, Connector, IPTypes]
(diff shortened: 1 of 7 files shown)


google/cloud/sql/connector/__init__.py
```diff
@@ -17,7 +17,7 @@
 from typing import List
 
 from .connector import connect, Connector
-from .instance_connection_manager import IPTypes
+from .instance import IPTypes
```
jackwotherspoon, line 20: @kurtisvg is this a breaking change?

kurtisvg (approved):

google/cloud/sql/connector/__init__.py
kurtisvg: Good question. I don't think so, because I _think_ the user still imports it from the same module ( `from google.cloud.sql.connector import IPTypes`)

However, our documentation doesn't show this. Can you update it to include the imports: https://github.com/GoogleCloudPlatform/cloud-sql-python-connector#specifying-public-or-private-ip

jackwotherspoon: Created PR #317 to address this

PR #316, opened 2022-04-04
feat!: deprecate default connect method

jackwotherspoon:
We want to now recommend to users to initialize a `Connector` object and not rely on the default `connect` method to do this for them. Will deprecate the default `connect` method in a future release.

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -261,6 +261,13 @@ def connect(instance_connection_string: str, driver: str, **kwargs: Any) -> Any:
     :returns:
         A DB-API connection to the specified Cloud SQL instance.
     """
+    # deprecation warning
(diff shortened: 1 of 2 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -261,6 +261,13 @@ def connect(instance_connection_string: str, driver: str, **kwargs: Any) -> Any:
     :returns:
         A DB-API connection to the specified Cloud SQL instance.
     """
+    # deprecation warning
+    logger.warning(
+        "Default `connect` method is deprecated and will be removed in a later "
```
kurtisvg, line 266:
```suggestion
        "The global `connect` method is deprecated and may be removed in a later "
```

kurtisvg (changes requested):

PR #331, opened 2022-04-07
chore: update install instructions per driver

### LINKED ISSUE #330: pip install no matched found for cloud-sql-python-connector[pg8000]
## Bug Description

Following the guide in the README of this repo, I was trying to connect to a Google cloud SQL PostgreSQL instance through python 3.7 from my local machine. I wasn't able to install the module `cloud-sql-python-connector[pg8000]`, as it returned `zsh: no matches found: cloud-sql-python-connector[pg8000]` in my terminal. I tried for all the other database drivers listed, and they did not work either.

Any help would be greatly appreciated!

## Example code (or command)

```
pip install cloud-sql-python-connector[pg8000]
```

```python
import sqlalchemy
import pg8000

# initialize Connector object
connector = Connector()


# function to return the database connection
def getconn() -> pg8000.dbapi.Connection:
    conn: pg8000.dbapi.Connection = connector.connect(
        "project:region:instance",
        "pg8000",
        user="postgres",
        password="XXXXXXXX",
        db="database-name"
    )
    return conn


# create connection pool
pool = sqlalchemy.create_engine(
        "postgresql+pg8000://",
        creator=getconn,
)
```

## Environment

1. OS type and version: `macOS 11.6.4`
2. Python version: `3.7`
2. Connector version: `0.6.0`

jackwotherspoon:
Closes #330

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,23 @@ Currently Python versions >= 3.7 are supported.
 
 ## Installation
 
-You can install this library with `pip install`, replacing `driver` with one of the database driver names listed above:
-```
-pip install cloud-sql-python-connector[driver]
-```
-For example, to use the Python connector with `pymysql`, run `pip install cloud-sql-python-connector[pymysql]`
+You can install this library with `pip install`, specifying the driver
+based on your database dialect.
 
-**Note:** MacOS users with the **zsh** terminal, please use
-`pip install 'cloud-sql-python-connector[driver]'` to avoid issues.
+### MySQL
+```
+pip install "cloud-sql-python-connector[pymysql]"
+```
+
+### Postgres
+```
+pip install "cloud-sql-python-connector[pg8000]"
+```
+
+### SQL Server
+```
+pip install "cloud-sql-python-connector[pytds]"
+```
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,9 @@ pip install cloud-sql-python-connector[driver]
 ```
 For example, to use the Python connector with `pymysql`, run `pip install cloud-sql-python-connector[pymysql]`
 
+**Note:** MacOS users with the **zsh** terminal, please use
```
enocom, line 40: This is a problem in zsh on Linux as well. I think it's probably easiest to just put quotes around the library name for all users.

jackwotherspoon: That fails for Windows command prompt then. I thought of that too but trying it on Windows didn't work.

kurtisvg: Does double quotes work on windows `pip install "cloud-sql-python-connector[pymysql]"?

jackwotherspoon: Yup double quotes actually does seem to work on Windows

enocom: Let's update to double quotes in the example then.

Also, would it be worth just listing all the drivers and their corresponding pip command?

enocom (approved):

kurtisvg (approved):

PR #339, opened 2022-04-18
chore: improve test coverage and mocks

jackwotherspoon:
- Add mocks for certs and keys
- Consolidate mocks to `mocks.py` file
- Improve unit test coverage from ~70% --> 92%
- Update unit tests to use `Connector()` over global `connect()`


.mypy.ini
```diff
@@ -2,6 +2,7 @@
 python_version = 3.7
 warn_unused_configs = True
 plugins = sqlmypy
+exclude = tests/unit
```
kurtisvg, line 5: Why?

tests/conftest.py
```diff
@@ -97,3 +99,31 @@ def fake_credentials() -> Credentials:
...
+
+
+def mock_server(server_sock: socket.socket) -> None:
+    """Create mock server listening on specified ip_address and port."""
+    ip_address = "127.0.0.1"
+    port = 3307
+    server_sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    server_sock.bind((ip_address, port))
+    server_sock.listen(0)
+    server_sock.accept()
```
kurtisvg, line 111: Will this only accept a single connection? 

Should we be checking to make sure that the connection is valid and we can send content through it?

tests/unit/test_connector.py
```diff
@@ -79,25 +70,46 @@ def test_connect_enable_iam_auth_error() -> None:
...
 
-def test_default_Connector_Init() -> None:
-    """Test that default Connector __init__ sets properties properly."""
-    default_connector = connector.Connector()
+
+def test_global_connect() -> None:
+    """Test that global connect properly initializes a singleton Connector."""
+    connect_string = "test-project:test-region:test-instance"
+    # stub Connector.connect
+    setattr(Connector, "connect", stub_connect)
```
kurtisvg, line 102: Would it be better to stub the metadata and connect to the local mock server?

tests/unit/test_instance.py
```diff
@@ -15,38 +15,34 @@
...
-
-
-async def _get_metadata_expired(*args: Any, **kwargs: Any) -> MockMetadata:
-    return MockMetadata(datetime.datetime.now() - datetime.timedelta(minutes=10))
-
-
-async def _get_metadata_error(*args: Any, **kwargs: Any) -> None:
-    raise BadRefresh("something went wrong...")
+# import mocks
+from mocks import (
```
kurtisvg, line 34: Would it be better to just import mocks at this point?

tests/unit/test_instance.py
```diff
@@ -285,3 +281,162 @@ async def test_Instance_close(instance: Instance) -> None:
...
+) -> None:
+    """
+    Test that _perform_refresh returns valid InstanceMetadata object.
+    """
+    # generate keys and certs for test
+    cert, priv_key = generate_cert("my-project", "my-instance")
+    server_ca_cert = self_signed_cert(cert, priv_key)
+    keys = asyncio.create_task(get_keys())
+    _, client_key = await keys
+    ephemeral_cert = client_key_signed_cert(cert, priv_key, client_key)
```
kurtisvg, line 303: Rather than have to do this a bunch of times, would it be better to make a "FakeProxyServer" object or similar tha does it in one location?

For example, see what we did in the Go connector: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/e29f8c13583d3ddb65ebec907ccdd300643efd82/internal/mock/cloudsql.go#L146

It's a single object creation but guarantees we create a fake backend, including the API requests and proxy server.

kurtisvg (changes requested):

tests/unit/test_connector.py
jackwotherspoon: This test is really meant to only test that calling the global connector initializes a `Connector` object with the proper defaults. Wouldn't testing that the global connector can make it all the way to connect to a server, be more of an integration test?

tests/conftest.py
kurtisvg: it would maybe be nice to test that the connection works, but not necessarily required.

Once thing is that it's fine for there to be overlap between unit and integration tests, because the unit tests are run locally and so should try to catch more issues than we need the integration tests to do.

tests/unit/test_connector.py
kurtisvg: The line between unit and integration test is only defined by what the unit(s) are. In this case, I think it's better to consistently stub/mock at the same place, so that we can ensure that parts are all working together. Since we're mocking the actual service I wouldn't really consider it an integration test.

tests/unit/test_instance.py
```diff
@@ -211,13 +214,23 @@ async def test_force_refresh_cancels_pending_refresh(
     """
     Test that force_refresh cancels pending task if refresh_in_progress event is not set.
     """
+    # with patch("google.auth.default") as mock_auth:
```
kurtisvg, line 217: Should this be commented out?

kurtisvg (changes requested):

kurtisvg (approved):

PR #390, opened 2022-07-04
WIP: add support for asyncpg driver

### LINKED ISSUE #218: Add support for asyncpg driver.
Support PostgreSQL [asyncpg](https://github.com/MagicStack/asyncpg) async driver with connector.

jackwotherspoon:
Todo:
- [ ] Wait for 0.26.0 release of asyncpg
- [ ] Determine how to use creator argument or similar SQLAlchemy [create_async_engine](https://docs.sqlalchemy.org/en/14/dialects/postgresql.html?highlight=asyncpg#module-sqlalchemy.dialects.postgresql.asyncpg) 
- [ ] Update README with asyncpg connection example

Closes #218


google/cloud/sql/connector/asyncpg.py
```diff
@@ -0,0 +1,77 @@
...
+    return await asyncpg.connect(
+        user=user,
+        database=db,
+        password=passwd,
+        host=ip_address,
+        port=SERVER_PROXY_PORT,
+        ssl=ctx,
+        direct_tls=True,
+        **kwargs,
+    )
```
jackwotherspoon, line 77: Currently SQLAlchemy's [create_async_engine](https://docs.sqlalchemy.org/en/14/dialects/postgresql.html?highlight=asyncpg#module-sqlalchemy.dialects.postgresql.asyncpg) for asyncpg does not have an equivalent to the synchronous engine's `creator` argument. So currently to enable connection pooling I've allowed the python connector to either return an `asyncpg.Connection` object as well as a `asyncpg.Pool` connection pool if the keyword argument `pool=True`

jackwotherspoon: Looks like from this [discussion thread](https://github.com/sqlalchemy/sqlalchemy/discussions/7783) that one of the SQLAlchemy maintainers agrees that an `async_creator` parameter would be a good idea.  I'll see if anyone has raised it as an issue and raise it myself if not.

jackwotherspoon: Removed pool for now, will make it a separate PR if we decide to implement in future

README.md
```diff
@@ -275,6 +283,58 @@ connector.connect(
 )
 ``` 
 
+### Async Driver Usage
+The Cloud SQL Connector for Python currently supports the
+[asyncpg](https://magicstack.github.io/asyncpg) Postgres database driver.
+This driver leverages [asyncio](https://docs.python.org/3/library/asyncio.html)
+to improve the speed and efficiency of database connections through concurrency.
```
kurtisvg, line 290: nit: I would flip the phrasing of this and say "The Cloud SQL Connect is compatible with asyncio to improve the speed and efficiency of database connection through concurrency. You can use all of the non-asyncio drivers through the `Connector.connect_async` function, and the follow asyncio ones:

README.md
```diff
@@ -275,6 +283,58 @@ connector.connect(
...
+
+The Cloud SQL Connector has an async `create_async_connector` that can be used
```
kurtisvg, line 292: It's maybe worth calling out the exact behavior -- that using `create_async_connector` will create a Connector in the current loop, while Connector will create a new loop in a background thread for internal use. 

The 2nd might be preferable with the sync drivers in an async context.

README.md
```diff
@@ -275,6 +283,58 @@ connector.connect(
...
+
+async def main():
+    # intialize Connector object using 'create_async_connector'
+    connector = await create_async_connector()
+
+    # create connection to Cloud SQL database
+    conn: asyncpg.Connection = await connector.connect_async(
+        "project:region:instance", # Cloud SQL instance connection name
+        "asyncpg",
+        user="root",
```
kurtisvg, line 314: nit: not root

README.md
```diff
@@ -275,6 +283,58 @@ connector.connect(
...
+```python
+import asyncpg
+from google.cloud.sql.connector import create_async_connector
+
+async def main():
+    # intialize Connector object using 'create_async_connector'
+    connector = await create_async_connector()
+
+    # create connection to Cloud SQL database
+    conn: asyncpg.Connection = await connector.connect_async(
```
kurtisvg, line 311: nit: this connection should probably use an async with

tests/unit/mocks.py
```diff
@@ -238,3 +239,12 @@ def generate_ephemeral(self, client_bytes: str) -> str:
...
         )
+
+
+async def wait_for_keys(future: asyncio.Future) -> Tuple[bytes, str]:
+    """
+    Helper method to await keys of Connector in tests prior to
+    initializing an Instance object.
+    """
+    keys = await future
+    return keys
```
kurtisvg, line 250: Why?

kurtisvg (changes requested):

README.md
jackwotherspoon: the connector or connection itself? `async with create_async_connector()` or `async with connector.connect_async(...)` ?

kurtisvg: The connection itself. 

Do we support async enters and exit with the Connector?

tests/conftest.py
```diff
@@ -192,7 +193,9 @@ async def connector(fake_credentials: Credentials) -> AsyncGenerator[Connector,
         mock_auth.return_value = fake_credentials, None
         # mock Cloud SQL Admin API calls
         mock_instance = FakeCSQLInstance(project, region, instance_name)
-        _, client_key = connector._keys.result()
+        _, client_key = asyncio.run_coroutine_threadsafe(
+            wait_for_keys(connector._keys), connector._loop
+        ).result()
```
jackwotherspoon, line 198: @kurtisvg here is the reason for `wait_for_keys` helper function. 

In the `connector` fixture that creates a `Connector` and mocks the API calls, I need the client key in order to mock the ephemeral cert. Since we now call `wrap_future(keys)` in the Connector's __init__ I can no longer call `connector._keys.result()` since the keys are now an `asyncio.Future` instead of `concurrent.Future`. A concurrent future will block on `.result()` and return the result once it is finished (does not need to be awaited), however `.result()` on an asyncio future will throw an error if result is not set when it is called (needs to be await prior to calling .result). This helper function basically allows me to turn an `asyncio.Future` into a `concurrent.Future`, unwrapping the future so I can access its value and mock the API calls prior to creating an `Instance` object.

README.md
jackwotherspoon: not currently, we don't have an `__aenter__` or `__aexit__` implemented.

jackwotherspoon: I added the ability to create an async context manager with `Connector` and added example to README. I can't use `async with` for the connection because the [asyncpg.Connection](https://github.com/MagicStack/asyncpg/blob/master/asyncpg/connection.py#L42) class does not have an `__aenter__` or `__aexit__`. We could potentially create a `Connection` wrapper class of our own to implement an async context manager for the database connection. ([Example](https://stackoverflow.com/questions/62507653/python-context-manager-for-asyncpg)). However, if in a future PR we are going to support connection pools (either through SQLAlchemy or native `asyncpg.Pool` then they will have context manager built in so not sure it is worth it to create the wrapper class.

tests/conftest.py
kurtisvg: I'm not sure I follow. Why don't you just await the keys (like it's done in the test on L160)?

Even if we do need this, it seems like it should be an anonymous function here as opposed to something in mocks.

README.md
kurtisvg: Ok, this is surprising to me that they don't support those two functions. I think it's fine to leave for now (maybe worth asking why they haven't?)

kurtisvg (changes requested):

README.md
jackwotherspoon: Seems like there is an [open issue](https://github.com/MagicStack/asyncpg/issues/760) for it actually, if I have some spare time I will try and put up another PR to asyncpg and add the two functions.

tests/conftest.py
jackwotherspoon: I cleaned up L160 to just use a task, it works because the present event loop is running in the current thread and can be awaited. This is not the case because for the `connector` fixture. The `connector._loop` is running in the connector's background thread and can not be awaited from the fixture, thus the need to have the helper function await in the background thread so the result can be accessed within the fixture.

kurtisvg: Ok so to summarize, you are converting an awaitable (asyncio.Future) into a concurrent.Future. You are using a closure to convert your awaitable into a coro because `run_coroutine_threadsafe` only takes coros.

It is probably worth trying to leave a comment that tries to summarize that. 

Also, make sure you've ruled out using `asyncio.wait()` -- it looks like it returns a coro, so it might be used here.

kurtisvg (approved):

tests/conftest.py
jackwotherspoon: Added comments and ruled out `asyncio.wait()` by testing it.

PR #392, opened 2022-07-05
chore: update system tests to use Connector object

jackwotherspoon:
Updating system tests to use `Connector` object over global `connector.connect` method which is deprecated and will be removed in future release.

### DIFF
--- a/tests/system/test_ip_types.py
+++ b/tests/system/test_ip_types.py
@@ -20,12 +20,14 @@ import uuid
 import pymysql
 import pytest
 import sqlalchemy
-from google.cloud.sql.connector import connector, IPTypes
+from google.cloud.sql.connector import Connector, IPTypes
 
 table_name = f"books_{uuid.uuid4().hex}"
 
 
-def init_connection_engine(ip_type: IPTypes) -> sqlalchemy.engine.Engine:
+def init_connection_engine(
+    connector: Connector, ip_type: IPTypes
+) -> sqlalchemy.engine.Engine:
     def getconn() -> pymysql.connections.Connection:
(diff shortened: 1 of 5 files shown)


tests/system/test_ip_types.py
```diff
@@ -45,19 +47,21 @@ def getconn() -> pymysql.connections.Connection:
...
-        pool = init_connection_engine(IPTypes.PUBLIC)
-    except Exception as e:
-        logging.exception("Failed to initialize pool with public IP", e)
-    with pool.connect() as conn:
-        conn.execute("SELECT 1")
+    with Connector() as connector:
+        try:
+            pool = init_connection_engine(connector, IPTypes.PUBLIC)
+        except Exception as e:
+            logging.exception("Failed to initialize pool with public IP", e)
```
kurtisvg, line 54: Should we allow this exception to bubble up? Catching it means getting a None exception later instead.

tests/system/test_ip_types.py
```diff
@@ -45,19 +47,21 @@ def getconn() -> pymysql.connections.Connection:
...
-        pool = init_connection_engine(IPTypes.PRIVATE)
-    except Exception as e:
-        logging.exception("Failed to initialize pool with private IP", e)
-    with pool.connect() as conn:
-        conn.execute("SELECT 1")
+    with Connector() as connector:
+        try:
+            pool = init_connection_engine(connector, IPTypes.PRIVATE)
+        except Exception as e:
+            logging.exception("Failed to initialize pool with private IP", e)
```
kurtisvg, line 65: Same here.

kurtisvg (approved):

PR #410, opened 2022-07-26
fix: stop event loop on Connector.close

### LINKED ISSUE #409: Asyncio loop not cleared on closing / garbage collection
## Bug Description

We have a use case where we create connectors on demand from users to connect to Cloud SQL. This implies that we create a `Connector` instance and then `close()` it. After that, the reference to the connector no longer exists so Python GC it.

The issue we see is the exception `OSError`: `[Errno 24] Too many open files` raised at the creation of a connector after a long time. Especially, it is when the asyncio loop is created (see below the stacktrace).

## Example code

```python
with Connector(credentials=credentials) as connector:
    try:
        yield connector.connect(
            self.authentication.instance,
            "pg8000",
            user=self.authentication.username,
            password=self.authentication.password,
            db=self.authentication.database,
        )
        break
    except GoogleAPICallError:
        raise IssueException()
    except ClientResponseError as e:
        if e.status == 429:
            time.sleep(5)
            continue
        else:
            raise IssueException()

```

## Stacktrace
```
OSError: [Errno 24] Too many open files
  File "salescrew/core/celery.py", line 146, in __call__
    raise e
  File "salescrew/core/celery.py", line 143, in __call__
    self.run(*args, **kwargs)
  File "salescrew/connectors/routines.py", line 153, in update_users_from_integrations
    integration_user = connector.get_user(user_integration.id)
  File "salescrew/connectors/databases/postgresql.py", line 297, in get_user
    with self.connection as conn:
  File "contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "salescrew/connectors/databases/postgresql.py", line 346, in connection
    connector = Connector(credentials=credentials)
  File "/usr/local/lib/python3.10/site-packages/google/cloud/sql/connector/connector.py", line 67, in __init__
    self._loop: asyncio.AbstractEventLoop = asyncio.new_event_loop()
  File "asyncio/events.py", line 782, in new_event_loop
    return get_event_loop_policy().new_event_loop()
  File "asyncio/events.py", line 673, in new_event_loop
    return self._loop_factory()
  File "asyncio/unix_events.py", line 64, in __init__
    super().__init__(selector)
  File "asyncio/selector_events.py", line 56, in __init__
    self._make_self_pipe()
  File "asyncio/selector_events.py", line 103, in _make_self_pipe
    self._ssock, self._csock = socket.socketpair()
  File "socket.py", line 607, in socketpair
    a, b = _socket.socketpair(family, type, proto)
```

## How to reproduce

  1. Create a lot of connectors one after another

## Environment

1. OS type and version:  Linux 5.15.55-1-lts
2. Python version: Python 3.10.4
2. Connector version: 0.6.2

## Proposal of resolution

I think because the connector uses the `run_forever` method of the loop, it doesn't clear itself after. Maybe a `self._loop.close()` might need to be done on the closing of the connector ? Or because the loop is used in a thread, we might also need to delete the thread at some point ?

jackwotherspoon:
Background threads used for `Connector()` are daemon threads and by default do not get killed until process exits. Since we are spinning up these threads with `loop.run_forever` the thread will not get killed until the event loop is closed. Which in the current state is when the Python program exits, which is not efficient for long running programs.

To fix this, we should call `loop.stop` when closing a `Connector` object which will in turn kill the background daemon thread.

Closes #409

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -284,8 +284,12 @@ class Connector:
         )
         # Will attempt to safely shut down tasks for 5s
         close_future.result(timeout=5)
-        # stop event loop (killing daemon thread)
-        self._loop.call_soon_threadsafe(self._loop.stop)
+        # if background thread exists for Connector, clean it up
+        if self._thread:
+            # stop event loop running in background thread
+            self._loop.call_soon_threadsafe(self._loop.stop)
+            # wait for thread to finish closing (i.e. loop to stop)
+            self._thread.join()
 
     async def close_async(self) -> None:
         """Helper function to cancel Instances' tasks
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -284,6 +284,9 @@ class Connector:
         )
         # Will attempt to safely shut down tasks for 5s
         close_future.result(timeout=5)
+        # if background thread, stop event loop (killing daemon thread)
+        if self._thread:
+            self._loop.call_soon_threadsafe(self._loop.stop)
 
     async def close_async(self) -> None:
         """Helper function to cancel Instances' tasks
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -284,9 +284,8 @@ class Connector:
         )
         # Will attempt to safely shut down tasks for 5s
         close_future.result(timeout=5)
-        # if background thread, stop event loop (killing daemon thread)
-        if self._thread:
-            self._loop.call_soon_threadsafe(self._loop.stop)
+        # stop event loop (killing daemon thread)
+        self._loop.call_soon_threadsafe(self._loop.stop)
 
     async def close_async(self) -> None:
         """Helper function to cancel Instances' tasks
--- a/tests/unit/test_connector.py
+++ b/tests/unit/test_connector.py
@@ -16,6 +16,7 @@ limitations under the License.
 
 import pytest  # noqa F401 Needed to run the tests
 import asyncio
+import threading
 
 from google.cloud.sql.connector import Connector, IPTypes, create_async_connector
 from google.cloud.sql.connector.connector import ConnectorLoopError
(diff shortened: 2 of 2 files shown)


tests/unit/test_connector.py
```diff
@@ -157,3 +157,16 @@ async def test_create_async_connector() -> None:
...
     assert connector._loop == asyncio.get_running_loop()
     await connector.close_async()
+
+
+def test_Connector_close_kills_threads() -> None:
+    """Test that Connector.close kills background threads."""
+    # several fixtures use threads so get count prior to test
+    prior_threads = threading.active_count()
+    # open and close 10 Connector objects
+    for _ in range(10):
```
kurtisvg, line 167: Is it possible to access the thread directly and call `is_alive()` after the close is called? This seems like a more reliable way to determine if it's still running or not

google/cloud/sql/connector/connector.py
```diff
@@ -284,6 +284,8 @@ def close(self) -> None:
         )
         # Will attempt to safely shut down tasks for 5s
         close_future.result(timeout=5)
+        # stop event loop (killing daemon thread)
+        self._loop.call_soon_threadsafe(self._loop.stop)
```
kurtisvg, line 288: 1. I don't think we want to close the loop when it's been handed in by someone else (i.e. such as in the Instance constructor or in async land) We should probably only close if Thread=None. Also, we might want to exit the thread as well? 

2. I think because we are scheduling with call_soon_threadsafe that this hasn't actually happened by the time `close` has exited. Are we sure this is the behavior we want? Since we are calling.result above, I think that the implicit contract might be that when close() ends, the resources are cleaned up.

kurtisvg (changes requested):

tests/unit/test_connector.py
jackwotherspoon: Done

google/cloud/sql/connector/connector.py
jackwotherspoon: How does this look? Looped over 100 `with Connector():` lines and only 1 thread was ever running at a time:
```python
# if background thread exists for Connector, clean it up
if self._thread:
    # stop event loop running in background thread
    self._loop.call_soon_threadsafe(self._loop.stop)
    # wait for thread to finish closing (i.e. loop to stop)
    self._thread.join()
```

kurtisvg (approved):

PR #424, opened 2022-08-10
chore: prefer TLSv1.3 for all connections

### LINKED ISSUE #418: Prefer TLS 1.3
When possible, we should prefer TLS 1.3, or at least verify that the connector already has this behavior.

jackwotherspoon:
Prefer TLSv1.3 and warn users if falling back to TLS v1.2

fixes #418

### DIFF
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -116,8 +116,14 @@ class InstanceMetadata:
     ) -> None:
         self.ip_addrs = ip_addrs
 
+        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+
         # verify OpenSSL version supports TLSv1.3
-        if not ssl.HAS_TLSv1_3:
+        if ssl.HAS_TLSv1_3:
+            # set minimum TLS version to TLSv1.3 if supported by client
+            self.context.minimum_version = ssl.TLSVersion.TLSv1_3
+        # fallback to TLSv1.2 for older versions of OpenSSL
+        else:
             if enable_iam_auth:
                 raise TLSVersionError(
                     f"Your current version of OpenSSL ({ssl.OPENSSL_VERSION}) does not "
@@ -129,12 +135,11 @@ class InstanceMetadata:
                 f"({ssl.OPENSSL_VERSION}), falling back to TLSv1.2\n"
                 "Upgrade your OpenSSL version to 1.1.1 for TLSv1.3 support."
             )
+            # set minimum TLS version to TLSv1.2
+            self.context.minimum_version = ssl.TLSVersion.TLSv1_2
 
-        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS)
         # add request_ssl attribute to ssl.SSLContext, required for pg8000 driver
         self.context.request_ssl = False  # type: ignore
-        # set minimum TLS version to TLSv1.2
-        self.context.minimum_version = ssl.TLSVersion.TLSv1_2
 
         self.expiration = expiration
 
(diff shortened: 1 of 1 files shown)


google/cloud/sql/connector/instance.py
```diff
@@ -126,13 +116,26 @@ def __init__(
...
+                f"({ssl.OPENSSL_VERSION}), falling back to TLSv1.2\n"
+                "Upgrade your OpenSSL version to 1.1.1 for TLSv1.3 support."
             )
 
-        self.context = ConnectionSSLContext()
+        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+        # add request_ssl attribute to ssl.SSLContext, required for pg8000 driver
+        self.context.request_ssl = False  # type: ignore
+        # set minimum TLS version to TLSv1.2
+        self.context.minimum_version = ssl.TLSVersion.TLSv1_2
```
enocom, line 137: Should we be setting minimum to TLS 1.3 and falling back to 1.2 when we can’t?

jackwotherspoon: I don't think it is necessary in my opinion, `ssl.SSLContext(ssl.PROTOCOL_TLS)` will use TLS 1.3 if available based on the documentation for `ssl.PROTOCOL_TLS`:
> ssl.PROTOCOL_TLS: Selects the highest protocol version that both the client and server support.

enocom: This might be a pedantic nit, but what about this?

``` python
if ssl.HAS_TLSv1_3:
    # set minimum TLS version to TLSv1.3
    self.context.minimum_version = ssl.TLSVersion.TLSv1_3
else:
    if enable_iam_auth:
        raise TLSVersionError(
            f"Your current version of OpenSSL ({ssl.OPENSSL_VERSION}) does not "
            "support TLSv1.3, which is required to use IAM Authentication.\n"
            "Upgrade your OpenSSL version to 1.1.1 for TLSv1.3 support."
        )
    logger.warning(
        "TLSv1.3 is not supported with your version of OpenSSL "
        f"({ssl.OPENSSL_VERSION}), falling back to TLSv1.2\n"
        "Upgrade your OpenSSL version to 1.1.1 for TLSv1.3 support."
    )
    # set minimum TLS version to TLSv1.2
    self.context.minimum_version = ssl.TLSVersion.TLSv1_2

self.context = ConnectionSSLContext()
self.context = ssl.SSLContext(ssl.PROTOCOL_TLS)
# add request_ssl attribute to ssl.SSLContext, required for pg8000 driver
self.context.request_ssl = False  # type: ignore
```

kurtisvg: I agree with Eno that we should use the min as 1.3 if the client supports it. The server side proxy should always support 1.3, so making it required when we know the client supports it will help prevent attacks where the attacker might force a downgrade

jackwotherspoon: Done

enocom (approved):

google/cloud/sql/connector/instance.py
```diff
@@ -126,13 +116,31 @@ def __init__(
...
+                    f"Your current version of OpenSSL ({ssl.OPENSSL_VERSION}) does not "
+                    "support TLSv1.3, which is required to use IAM Authentication.\n"
+                    "Upgrade your OpenSSL version to 1.1.1 for TLSv1.3 support."
+                )
+            logger.warning(
+                "TLSv1.3 is not supported with your version of OpenSSL "
+                f"({ssl.OPENSSL_VERSION}), falling back to TLSv1.2\n"
+                "Upgrade your OpenSSL version to 1.1.1 for TLSv1.3 support."
             )
+            # set minimum TLS version to TLSv1.2
```
kurtisvg, line 138: super-nit: redundant comment

kurtisvg (approved):

Issue #310, opened 2022-03-26
Make user optional when passing enable_iam_auth


aebrahim:
## Feature Description

If I'm using this in an environment like cloud run, there really only is one account - the currently active service account. I think the default assumption is that the user that is going to be doing the proxying is also the one that is being logged in, so it's quite redundant to have to get the current service account, and then chop off the `gserviceaccount.com` as required.

A cleaner API would be to simply no longer have user required when enable_iam_auth is passed, and have the connector correctly detect service account emails and appropriately chop them, cleanly abstracting away the internals.

That would take the common use case from
```
account = detect_account()
if account.endswith(".iam.gserviceaccount.com"): 
    account = account.replace(".gserviceaccount.com", "")
connector.connect(MY_DB_STR, "pg8000", user=account, db=MY_DB, enable_iam_auth=True)
```
to the extremely elegant `connector.connect(MY_DB_STR, "pg8000", db=MY_DB, enable_iam_auth=True)`

## Alternatives Considered
We could continue to keep this redundant, but it's forcing the calling code to still be dealing with GCP internals.

Thank you for your consideration :P

jackwotherspoon:
Hi @aebrahim, thanks so much for raising the feature request! We appreciate you taking the time in helping to better the Python Connector :)

This is a very interesting idea that I definitely see the benefit in having. Not only would it make it easier to use IAM AuthN but it would also allow for us to abstract away certain database specific IAM email considerations (Postgres removes the `.gserviceaccount.com`, MySQL removes everything prior to the `@` symbol)  allowing the user to interact with each in the same way.

However, implementation wise this is slightly trickier to achieve. The default credentials object used by the connector for authentication don't always have an email attribute associated with them. It does for service account credentials (so would work for Cloud Run, Cloud Functions, or when someone uses a service account key directly) but for credentials sourced from users logged in via gcloud etc. there is not an email attribute. 

I will discuss with the rest of the team if the service account use-case justifies the feature and how we would handle the user case.

I will keep you posted if the feature begins to be worked on and if I have any further questions regarding the feature request.

Thanks again for raising this and have a great day!

- Jack

pataquets:
@jackwotherspoon : Any news? Would it be acceptable to start supporting this under the most common scenarios, on straightforward environments such as Cloud Run?

PR #449, opened 2022-09-06
feat: format `user` argument for automatic IAM authn

### LINKED ISSUE #429: Validate email address format when using automatic IAM authentication
When adding IAM database users to either MySQL or Postgres the user or service account username is formatted in a certain way that differs across database engine types.

This is not necessarily trivial to users when using the connectors, we should try validating and formatting the IAM database usernames within the connectors to remove this burden on users.

**Postgres**: Removes `.gserviceaccount.com` domain suffix if it exists.
**MySQL**: Removes everything after and including the `@` sign. (`test-user@gmail.com` -> `test-user`)

Should already have `databaseVersion` from the API request to https://cloud.google.com/sql/docs/mysql/admin-api/rest/v1beta4/connect/get

jackwotherspoon:
Adding a formatting step to format the database `user` argument when using automatic IAM database authentication. This will abstract away the different formatting semantics between MySQL and Postgres so that users can leverage the connector without needing to worry about learning naming rules.

**Postgres**: Removes `.gserviceaccount.com` domain suffix if it exists.
**MySQL**: Removes everything after and including the `@` sign. (`test-user@gmail.com` -> `test-user`)

Closes #429


kurtisvg (changes requested):

tests/system/test_connector_object.py
```diff
@@ -24,6 +25,11 @@
 import concurrent.futures
 from threading import Thread
 
+from google.cloud.sql.connector.utils import (
+    InvalidPostgresDatabaseUser,
+    InvalidMySQLDatabaseUser,
+)
```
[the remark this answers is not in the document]
jackwotherspoon: done

google/cloud/sql/connector/connector.py
```diff
@@ -234,6 +234,11 @@ async def connect_async(
         # helper function to wrap in timeout
         async def get_connection() -> Any:
             instance_data, ip_address = await instance.connect_info(ip_type)
+
+            # validate `user` param for automatic IAM database authn
+            if enable_iam_auth:
+                validate_database_user(instance_data.database_version, kwargs["user"])
```
aebrahim, line 240: Drive by comment from a user of this library - hope y'all don't mind.

It would be really nice if `enable_iam_auth` is set, we do only a little bit of extra work to make #310 happen and format the user for the caller.

Sample code example
```python
if enable_iam_auth:
  if "user" not in kwargs:
    if not hasattr(self._credentials, "service_account_email"):
      # Tell the caller to either explicitly pass in user, or pass in credentials auth to service account.
      raise SomeError("some message")
    user = format_user_for_sql(instance_data.database_version, self._credentials.service_account_email)
  else:
    user = kwargs["user"]
    validate_database_user(instance_data.database_version, user)
```

aebrahim: I just figured if all the work is already being done to figure out what is appropriate for each dialect, it would be nice to just to it for the caller, so they don't have to duplicate this logic on their end.

jackwotherspoon: Hi @aebrahim, thanks for the comment! We always welcome ideas and suggestions for the connector 😄  

This PR in general is meant to be a step in the right direction towards what you have outlined here in the comment and was inspired by your original #310. I think it would be an awesome feature to have!

I've been looking into different ways to make the `user` optional. The main reason we aren't ready to pull the trigger on an implementation like the one you proposed just yet is because we want to maintain feature parity. Not only across different credential types, (i.e. service account vs. user credentials vs. application credentials) but also across our connector libraries. We try to maintain feature parity across this library, the [Go connector](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector) and the [Java connector](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory). While the feature is fairly straightforward to implement here in Python, the other languages aren't quite as easy.

We want to make the user experience as consistent as possible across the different languages. I will bring this back up with the team to discuss if diverging slightly here and offering this feature only in Python can be done, because I do see the immense value and improved user experience of the feature.

Thanks again for your enthusiasm and suggestion, they are always appreciated. I will hopefully have an update soon for you on this.

aebrahim: Thanks for the detailed reasoning (as well as all the hard work on this package) @jackwotherspoon!

jackwotherspoon:
@kurtisvg I've updated this PR to format the `user` argument as discussed offline. If you could take another look that would be great! 😄

google/cloud/sql/connector/utils.py
```diff
@@ -77,3 +77,33 @@ def write_to_file(
         priv_out.write(priv_key)
 
     return (ca_filename, cert_filename, key_filename)
+
+
+def remove_suffix(input_string: str, suffix: str) -> str:
```
kurtisvg, line 82: nit: I think you could just inline this function. It's only used the one place.

kurtisvg (approved):

google/cloud/sql/connector/connector.py
jackwotherspoon: @aebrahim based on your comment here we have updated the PR to actually format the `user` arg and not just validate it. This will abstract away the need to format MySQL and Postgres IAM users differently and allow developers to just pass in the full user or service account email as the `user` arg when using the Connector with IAM authentication.

This is hopefully a better step in the right direction to making the `user` arg optional down the road. Will try and prioritize this in the new year!

Thanks for the wonderful suggestions and feedback!

jackwotherspoon:
Code coverage dropped by `0.2%` which is the reason the check is failing. New missing coverage lines are the added debug line and the if statement surrounding it. These are not critical pieces of code to test and thus the failing coverage check is okay.

Issue #442, opened 2022-08-31
How do I debug an error using IAM authentication?


garimasingh2016:
## Question
I am trying to connect to a PostgresSQL database using a service account and automatic IAM authentication. I have followed all the steps in "Before you begin" [here](https://cloud.google.com/sql/docs/postgres/iam-logins#cloud-sql-auth-proxy) and am using the Python connector with the `asyncpg` driver.

Additionally I have run `gcloud auth activate-service-account` and can see that the current account the Gcloud CLI is using is the service account. However, I still get an authentication error and it's pretty undescriptive. I have no idea how to proceed or how to debug. Most likely, it is something very simple but I see very little documentation online or methods for troubleshooting. The code is below (I've changed the exact names of the database, instance, and service account for security):
```
def query_postgres(query, params):
    with Connector() as connector:
        cursor = connector.connect(
            "instance_name",
            "asyncpg",
            user="service_account@project_name.iam",
            db="database_name",
            enable_iam_auth=True,
        ).cursor()
        cursor.execute(query, params)
        rows = cursor.fetchall()
        return rows
```

This is the error:
```
asyncpg.exceptions.InvalidAuthorizationSpecificationError: Cloud SQL IAM service account authentication failed for user "service_account@project_name.iam"
```
Would appreciate if someone could give me some pointers for debugging!

jackwotherspoon:
Hi @garimasingh2016, thanks for raising a question and using the Cloud SQL Python Connector!

I can help give you some guidance and hopefully help you debug this! 

First off, if you are using the `asyncpg` driver you will want to follow the [guidance here](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector#async-driver-usage) to connect in a way that supports `asyncio` and asynchronous connections. It is slightly different than the synchronous `pg8000` driver. 

You will want to update your code to look something like:
```python
# add extra import
import asyncio

# make function async
async def query_postgres(query, params):
    # get current running event loop to be used with Connector
    loop = asyncio.get_running_loop()
    # intialize Connector object as async context manager
    async with Connector(loop=loop) as connector:
        # connect with connector.connect_async instead
        conn = await connector.connect_async(
            "instance_name",
            "asyncpg",
            user="service_account@project_name.iam",
            db="database_name",
            enable_iam_auth=True,
        )
```
[...]

PR #450, opened 2022-09-06
chore: update ADC details in README

jackwotherspoon:
Updating link to point to more specific header for details.

Based on feedback from #442

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,21 +88,60 @@ This library uses the [Application Default Credentials (ADC)][adc] strategy for
 resolving credentials. Please see the [google.auth][google-auth] package 
 documentation for more information on how these credentials are sourced.
 
+If you are using the Python Connector with Cloud Run, App Engine, Cloud Functions,
+or Compute Engine, the application default credentials will **automatically default**
+**to the service account used to run each service**.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,21 +88,35 @@ This library uses the [Application Default Credentials (ADC)][adc] strategy for
...
+
+To explicitly set a specific source for the credentials, see
 [Configuring the Connector](#configuring-the-connector) below.
 
-[adc]: https://cloud.google.com/docs/authentication
+[adc]: https://cloud.google.com/docs/authentication#adc
 [google-auth]: https://google-auth.readthedocs.io/en/master/reference/google.auth.html
 [cloud-sdk]: https://cloud.google.com/sdk/docs/install
+[sa-impersonation]: https://cloud.google.com/iam/docs/impersonating-service-accounts
+[sa-keys]: https://cloud.google.com/docs/authentication/provide-credentials-adc#local-key
```
kurtisvg, line 119: I think this is information that is not really python-connector specific and ideally should come from an official docs page that we can reference.

kurtisvg (changes requested):

kurtisvg (approved):

PR #466, opened 2022-09-12
feat: add support for MySQL auto IAM AuthN

jackwotherspoon:
Todo:
- [x] Add supporting test infra
- [x] <s>Update README</s> Will update once MySQL auto IAM authn is 100% rolled out


google/cloud/sql/connector/instance.py
```diff
@@ -326,11 +327,11 @@ async def _perform_refresh(self) -> InstanceMetadata:
             try:
                 metadata = await metadata_task
                 # check if automatic IAM database authn is supported for database engine
-                if self._enable_iam_auth and not metadata[
-                    "database_version"
-                ].startswith("POSTGRES"):
+                if self._enable_iam_auth and metadata["database_version"].startswith(
+                    "SQLSERVER"
```
kurtisvg, line 331: This worries me because it doesn't capture any future database versions that might be launched that don't start with SQLServer

kurtisvg (changes requested):

google/cloud/sql/connector/instance.py
jackwotherspoon: Good catch! Updated logic to be `not metadata["database_version"].startswith(("POSTGRES", "MYSQL"))` to account for this.

kurtisvg (approved):

PR #476, opened 2022-09-15
fix: throw error if Auto IAM AuthN is unsupported

### LINKED ISSUE #473: Connector should error on invalid IAM AuthN calls
If a database engine doesn't support Auto IAM AuthN (i.e., MySQL, SQL Server), the connector should throw an error if a user has requested the feature.

Related: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/issues/283.

jackwotherspoon:
Throw an error if `enable_iam_auth=True` is set for MySQL or SQL Server instances where automatic IAM database authentication is not currently supported.

Fixes #473


google/cloud/sql/connector/instance.py
```diff
@@ -356,10 +355,20 @@ async def _perform_refresh(self) -> InstanceMetadata:
                 )
             )
 
+            # gather Admin API tasks, return_exceptions=True to check auto iam authn
             metadata, ephemeral_cert = await asyncio.gather(
-                metadata_task, ephemeral_task
+                metadata_task, ephemeral_task, return_exceptions=True
```
kurtisvg, line 360: Why wouldn't we just let these exceptions percolate up?

google/cloud/sql/connector/instance.py
```diff
@@ -356,10 +355,20 @@ async def _perform_refresh(self) -> InstanceMetadata:
...
             )
 
+            if isinstance(metadata, BaseException):
+                raise metadata
+
+            # check if automatic IAM database authn is supported for database engine
+            verify_iam_auth(metadata["database_version"], self._enable_iam_auth)
```
kurtisvg, line 367: Why can't this check be inlined? 

We can just check here and raise exception:

```python
if self._enable_iam_auth and not supports_iam_auth(metadata["database_version"]):
     raise Exception()
```

kurtisvg (changes requested):

kurtisvg (approved):

kurtisvg (approved):

PR #488, opened 2022-09-28
feat: downscope token used for IAM DB AuthN

### LINKED ISSUE #486: Downscope OAuth2 token included in ephemeral certificate
The token included in the ephemeral cert should have just a https://www.googleapis.com/auth/sqlservice.login scope attached to it.

jackwotherspoon:
Down-scope credentials and OAuth2 token used in `generateEphemeral` API request for IAM AuthN to have only `"https://www.googleapis.com/auth/sqlservice.login"` scope.

Closes #486

### DIFF
--- a/google/cloud/sql/connector/refresh_utils.py
+++ b/google/cloud/sql/connector/refresh_utils.py
@@ -167,7 +167,7 @@ async def _get_ephemeral(
     elif not isinstance(pub_key, str):
         raise TypeError(f"pub_key must be of type str, got {type(pub_key)}")
 
-    if not credentials.valid or enable_iam_auth:
+    if not credentials.valid:
         request = google.auth.transport.requests.Request()
         credentials.refresh(request)
 
@@ -258,6 +258,8 @@ def _downscope_credentials(
     else:
         # create shallow copy to not overwrite scopes on default credentials
         scoped_creds = copy.copy(credentials)
+        # overwrite '_scopes' to down-scope user credentials
+        # Cloud SDK reference: https://github.com/google-cloud-sdk-unofficial/google-cloud-sdk/blob/93920ccb6d2cce0fe6d1ce841e9e33410551d66b/lib/googlecloudsdk/command_lib/sql/generate_login_token_util.py#L116
         scoped_creds._scopes = scopes
     # down-scoped credentials require refresh, are invalid after being re-scoped
     if not scoped_creds.valid:
--- a/google/cloud/sql/connector/refresh_utils.py
+++ b/google/cloud/sql/connector/refresh_utils.py
@@ -17,10 +17,11 @@ limitations under the License.
 
 import aiohttp
 import google.auth
-from google.auth.credentials import Credentials
+from google.auth.credentials import Credentials, Scoped
 import google.auth.transport.requests
-from typing import Any, Dict
+from typing import Any, Dict, List
 import datetime
+import copy
 import asyncio
 import logging
 
@@ -181,7 +182,8 @@ async def _get_ephemeral(
     data = {"public_key": pub_key}
 
     if enable_iam_auth:
-        data["access_token"] = credentials.token
+        login_creds = _downscope_credentials(credentials)
+        data["access_token"] = login_creds.token
 
     resp = await client_session.post(
         url, headers=headers, json=data, raise_for_status=True
@@ -229,3 +231,21 @@ async def _is_valid(task: asyncio.Task) -> bool:
         # supress any errors from task
         logger.debug("Current instance metadata is invalid.")
     return False
+
+
+def _downscope_credentials(
+    credentials: Credentials,
+    scopes: List[str] = ["https://www.googleapis.com/auth/sqlservice.login"],
+) -> Credentials:
+    if isinstance(credentials, Scoped):
+        scoped_creds = credentials.with_scopes(scopes=scopes)
+    # authenticated user credentials can not be re-scoped
+    else:
+        # create shallow copy to not overwrite scopes on default credentials
+        scoped_creds = copy.copy(credentials)
+        scoped_creds._scopes = scopes
+    # downscoped credentials require refresh
+    if not scoped_creds.valid:
+        request = google.auth.transport.requests.Request()
+        scoped_creds.refresh(request)
+    return scoped_creds
--- a/google/cloud/sql/connector/refresh_utils.py
+++ b/google/cloud/sql/connector/refresh_utils.py
@@ -237,6 +237,19 @@ def _downscope_credentials(
     credentials: Credentials,
     scopes: List[str] = ["https://www.googleapis.com/auth/sqlservice.login"],
 ) -> Credentials:
+    """Generate a down-scoped credential.
+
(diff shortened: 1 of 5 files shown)


enocom (approved):
:shipit:

shubha-rajan (approved):

google/cloud/sql/connector/refresh_utils.py
```diff
@@ -229,3 +231,34 @@ async def _is_valid(task: asyncio.Task) -> bool:
...
+    :param scopes
+        List of Google scopes to include in down-scoped credentials object.
+
+    :rtype: google.auth.credentials.Credentials
+    :returns: Down-scoped credentials object.
+    """
+    if isinstance(credentials, Scoped):
+        scoped_creds = credentials.with_scopes(scopes=scopes)
+    # authenticated user credentials can not be re-scoped
+    else:
```
kurtisvg, line 256: I'm not sure I follow this logic - why are we checking the type here?

google/cloud/sql/connector/refresh_utils.py
```diff
@@ -229,3 +231,34 @@ async def _is_valid(task: asyncio.Task) -> bool:
...
+        # create shallow copy to not overwrite scopes on default credentials
+        scoped_creds = copy.copy(credentials)
+        scoped_creds._scopes = scopes
```
kurtisvg, line 259: Is it safe to assume this type will always be here?

google/cloud/sql/connector/refresh_utils.py
```diff
@@ -229,3 +231,34 @@ async def _is_valid(task: asyncio.Task) -> bool:
         # supress any errors from task
         logger.debug("Current instance metadata is invalid.")
     return False
+
+
+def _downscope_credentials(
+    credentials: Credentials,
+    scopes: List[str] = ["https://www.googleapis.com/auth/sqlservice.login"],
```
kurtisvg, line 238: Do we scope to just `sqladmin` for the other API calls? (if not, probably should start doing it)

google/cloud/sql/connector/refresh_utils.py
```diff
@@ -229,3 +231,34 @@ async def _is_valid(task: asyncio.Task) -> bool:
...
+        scoped_creds = credentials.with_scopes(scopes=scopes)
+    # authenticated user credentials can not be re-scoped
+    else:
+        # create shallow copy to not overwrite scopes on default credentials
+        scoped_creds = copy.copy(credentials)
+        scoped_creds._scopes = scopes
+    # downscoped credentials require refresh
+    if not scoped_creds.valid:
+        request = google.auth.transport.requests.Request()
+        scoped_creds.refresh(request)
```
kurtisvg, line 263: We should refresh every time before we create the ephemeral cert, so that it's duration is as long as possible. I think it might also be good to move this logic out of the function and into the ephemeral cert func for this reason.

kurtisvg (changes requested):

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: Credentials sourced from service accounts and metadata can be re-scoped (are child class from `Scoped`), but credentials from an authenticated user can not.

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: We already do refresh before ephemeral cert API call ([code snippet](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/refresh_utils.py#L169-L171)). The reason for this refresh is that once you re-scope a credential and create a new credential object, it seems to become invalid and thus requires a new refresh.

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: The default credentials are created with the following scopes within [`_auth_init`](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/instance.py#L264-L267):
```python
scopes = [
    "https://www.googleapis.com/auth/sqlservice.admin",
    "https://www.googleapis.com/auth/cloud-platform",
]
```

Are you saying we may only need `"https://www.googleapis.com/auth/sqlservice.admin"` and not the second scope?

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: Which type? The `_scopes` argument?

google/cloud/sql/connector/refresh_utils.py
kurtisvg: Yes, we shouldn't need cloud-platform

google/cloud/sql/connector/refresh_utils.py
kurtisvg: Right - technically this is an internal value so I'm not sure if it's safe to reference. Can you link to the Cloud SDK example of them using it?

google/cloud/sql/connector/refresh_utils.py
kurtisvg: we shouldn't need to refresh here if it refreshes there then?

kurtisvg (changes requested):

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: I think we need to refresh both credentials don't we? We need to still refresh the default credential so its token is valid in the header, we also need to refresh the down-scoped credential for it's token to be valid. Thoughts?

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: removed the scope

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: added a comment in code referencing Cloud SDK implementation

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: I understand, removed the `or enable_iam_auth` from the if statement

google/cloud/sql/connector/refresh_utils.py
```diff
@@ -181,7 +182,8 @@ async def _get_ephemeral(
     data = {"public_key": pub_key}
 
     if enable_iam_auth:
-        data["access_token"] = credentials.token
+        login_creds = _downscope_credentials(credentials)
```
kurtisvg, line 185: Can we add a comment here that says "downscope credentials refreshes them, so we don't need to do so here". I want to make sure it's not lost in terms of importance to create_ephemeral

kurtisvg (approved):

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: done

PR #511, opened 2022-10-19
chore: require type hints

enocom:


enocom:
Should we update our config file to match something like this? https://blog.wolt.com/engineering/2021/09/30/professional-grade-mypy-configuration/

enocom:
@jackwotherspoon I'm going to close this. I think we might be able to make our type checking more strict here, but I'll leave that decision up to you.

PR #540, opened 2022-11-10
docs: add missing ref to asyncpg

ruyadorno:
Was reading through the source code and found this, which seems to be a simple missing reference compared with the comments elsewhere 😊

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -125,7 +125,7 @@ class Connector:
         :type driver: str
         :param: driver:
             A string representing the driver to connect with. Supported drivers are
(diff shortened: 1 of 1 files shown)


jackwotherspoon:
@ruyadorno the `connect` method is only called for synchronous drivers. For async drivers, they call [`connect_async`](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/connector.py#L155) directly. That is why I don't list asyncpg as supported with the `connect` method. Maybe that could be more clear somehow though.

ruyadorno:
@jackwotherspoon makes sense! I guess I'll just close this one for now 😊 thanks!

PR #575, opened 2023-01-04
fix: update TLS protocol for python3.10

jackwotherspoon:
In Python 3.10, `ssl.PROTOCOL_TLS` is deprecated and outputs warnings in logs/tests:
```
DeprecationWarning: ssl.PROTOCOL_TLS is deprecated
  self.context = ssl.SSLContext(ssl.PROTOCOL_TLS)
```
 Instead, one of [`ssl.PROTOCOL_TLS_CLIENT`](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT) or [`ssl.PROTOCOL_TLS_SERVER`](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_SERVER) should be used, in our case CLIENT.

However, `ssl.PROTOCOL_TLS_CLIENT` brings with it some [new changes](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT). It sets `check_hostname` to True and `verify_mode` to `CERT_REQUIRED` by default. (`PROTOCOL_TLS` defaulted these values to False and CERT_NONE)

These new defaults lead to an error when using the TLS connection:
```
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: IP address mismatch, certificate is not valid for 'XXX.XXX.XX.XX'.
```

Because of this I have added changing the defaults into the current PR.

### DIFF
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -79,7 +79,11 @@ class InstanceMetadata:
     ) -> None:
         self.ip_addrs = ip_addrs
         self.database_version = database_version
-        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+
+        # update ssl.PROTOCOL_TLS_CLIENT defaults
+        self.context.check_hostname = False
+        self.context.verify_mode = ssl.VerifyMode.CERT_NONE
 
         # verify OpenSSL version supports TLSv1.3
         if ssl.HAS_TLSv1_3:
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -81,9 +81,8 @@ class InstanceMetadata:
         self.database_version = database_version
         self.context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
 
-        # update ssl.PROTOCOL_TLS_CLIENT defaults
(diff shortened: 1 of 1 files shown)


google/cloud/sql/connector/instance.py
```diff
@@ -79,7 +79,11 @@ def __init__(
     ) -> None:
         self.ip_addrs = ip_addrs
         self.database_version = database_version
-        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+
+        # update ssl.PROTOCOL_TLS_CLIENT defaults
+        self.context.check_hostname = False
+        self.context.verify_mode = ssl.VerifyMode.CERT_NONE
```
kurtisvg, line 86: I don't follow why this is CERT_NONE -- the docs say:

> In this mode, certificates are required from the other side of the socket connection; an [SSLError](https://docs.python.org/3/library/ssl.html#ssl.SSLError) will be raised if no certificate is provided, or if its validation fails.

I think we definitely want CERT_REQURIED?

kurtisvg (changes requested):

google/cloud/sql/connector/instance.py
jackwotherspoon: Yup we do want `CERT_REQUIRED`, updated the PR to only change `check_hostname`.

kurtisvg (approved):

PR #589, opened 2023-01-16
chore: move `_auth_init` out of Instance class

jackwotherspoon:
The `_auth_init` method to initialize credentials does not need to be an `Instance` class method. It does not rely on any class attributes and thus can me moved to the`utils.py` file as a utility method.


shubha-rajan (approved):
LGTM!

PR #599, opened 2023-02-03
feat: improve reliability of certificate refresh

jackwotherspoon:
Making refresh calculation more reliable and the same when connecting using both auto IAM AuthN or regular built-in database users.


shubha-rajan (approved):

shubha-rajan:
LGTM!

enocom (approved):

PR #658, opened 2023-03-21
feat: add support for HTTP proxies

jackwotherspoon:
Add support for [aiohttp.Clientsession to enable http proxies](https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support).

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -70,10 +70,15 @@ class Connector:
         creating new event loop on background thread.
 
     :type sqladmin_api_endpoint: str
-    :param sqladmin_api_endpoint:
+    :param sqladmin_api_endpoint
         Base URL to use when calling the Cloud SQL Admin API endpoint.
         Defaults to "https://sqladmin.googleapis.com", this argument should
         only be used in development.
+
+    :type http_proxy: bool
+    :param http_proxy
+        Enables extracting proxy configuration from HTTP_PROXY, HTTPS_PROXY,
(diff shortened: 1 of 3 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -70,10 +70,15 @@ class Connector:
...
-    :param sqladmin_api_endpoint:
+    :param sqladmin_api_endpoint
         Base URL to use when calling the Cloud SQL Admin API endpoint.
         Defaults to "https://sqladmin.googleapis.com", this argument should
         only be used in development.
+
+    :type http_proxy: bool
+    :param http_proxy
+        Enables extracting proxy configuration from HTTP_PROXY, HTTPS_PROXY,
+        WS_PROXY or WSS_PROXY environment variables (all are case insensitive).
```
ruyadorno, line 81: nit: I believe a `enable_http_proxy` param name might convey better the idea that this is just a flag to allow for reading from the env vars. What do you think? (mentioning because when I first saw `http_proxy` I was expecting it to take the hostname value directly)

ruyadorno (approved):
just a small optional suggestion on the arg name otherwise 👍 LGTM

jackwotherspoon:
Turns out that aiohttp does not support socks proxies. I will look to create a new PR to instead allow users to pass in a custom `aiohttp.ClientSession` client so users can leverage custom packages that add socks support such as [aiohttp-socks](https://pypi.org/project/aiohttp-socks/) while not having the Python Connector depend on these custom libraries.

PR #754, opened 2023-06-06
chore: Update blunderbuss.yml

enocom:


jackwotherspoon:
@enocom are you sure you want to do this? I am fine being the only one on the blunderbuss for now.

PR #756, opened 2023-06-08
chore: add TYPE_CHECKING for imports

jackwotherspoon:
Adding `if TYPE_CHECKING:` to improve runtime efficiency. `TYPE_CHECKING` is False at runtime, allowing imports  that are only used for type hints etc. to only be imported by 3rd party static type checkers (aka lint job etc).

See [Python official docs](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) for more details.


jackwotherspoon:
Coverage drop is insignificant

enocom (approved):

PR #758, opened 2023-06-12
docs: document SQLAlchemy async connection pooling

### LINKED ISSUE #422: Support connection pooling for async drivers
The Cloud SQL Python Connector recommends using our library with connection pooling (usually through [SQLAlchemy](https://www.sqlalchemy.org/) library)

However, currently SQLAlchemy's `create_async_engine` method does not allow the use of an asynchronous `creator` argument [#8215](https://github.com/sqlalchemy/sqlalchemy/issues/8215)

This makes it very difficult to use connection pooling with the Cloud SQL Python Connector's async drivers. Currently there is a workaround that is both confusing and not practical to recommend to our users:

```python
import asyncio
from sqlalchemy import text
from sqlalchemy.ext.asyncio import create_async_engine
from sqlalchemy.util import await_only

async def async_creator():
    loop = asyncio.get_running_loop()
    async with Connector(loop=loop) as connector:
        conn = await connector.connect_async(
            "project:region:instance", # Cloud SQL instance connection name"
            "asyncpg",
            user="my-db-user",
            password="my-db-password",
            db="my-db-name",
        )
        return conn

async def async_main():
    def adapted_creator():
        dbapi = engine.dialect.dbapi
        from sqlalchemy.dialects.postgresql.asyncpg import (
            AsyncAdapt_asyncpg_connection,
)
        return AsyncAdapt_asyncpg_connection(
            dbapi,
            await_only(async_creator()),
            prepared_statement_cache_size=100,
        )

    # create async connection pool with wrapped creator
    engine = create_async_engine(
        "postgresql+asyncpg://",
        echo=True,
        creator=adapted_creator,
    )
    
    # use connection pooling with Cloud SQL Python Connector
    async with engine.connect() as conn:
        query = await conn.execute(text("SELECT * from ratings"))
        results = query.fetchall()
        for row in results:
            print(row)


asyncio.run(async_main())
```

<strike>For this reason, we should look at the possibility of supporting and returning native connection pools from the Cloud SQL Python Connector for async drivers. (ex. [asyncpg.create_pool](https://magicstack.github.io/asyncpg/current/usage.html#connection-pools))

This will allow users to have the benefits of connection pooling without the need for the confusing SQLAlchemy workaround.

I would suggest a `connector.create_pool` method or something that has the exact same interface as `connector.connect_async`:
```python
    # intialize Connector object using 'create_async_connector'
    connector = await create_async_connector()

    # create connection pool to Cloud SQL database
    conn: asyncpg.Pool = await connector.create_pool(
        "project:region:instance", # Cloud SQL instance connection name
        "asyncpg",
        user="my-db-user",
        password="my-db-pass",
        db="my-db-name",
        # ... additional database driver or connection pool args 
    )
```
</strike>

See [below comment](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/issues/422#issuecomment-1511775865) as to reasoning for not supporting native connection pooling in this library.

jackwotherspoon:
As of SQLAlchemy v2.0.16 `async_creator` is now a supported argument for `create_async_engine` allowing the Cloud SQL Python Connector to support async connection pooling

Closes #422

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,15 +440,15 @@ from google.cloud.sql.connector import Connector, create_async_connector
 async def init_connection_pool(connector: Connector) -> AsyncEngine:
     # initialize Connector object for connections to Cloud SQL
     async def getconn() -> asyncpg.Connection:
-            conn: asyncpg.Connection = await connector.connect_async(
-                "project:region:instance",  # Cloud SQL instance connection name
-                "asyncpg",
-                user="my-user",
-                password="my-password",
-                db="my-db-name"
-                # ... additional database driver args
-            )
-            return conn
+        conn: asyncpg.Connection = await connector.connect_async(
+            "project:region:instance",  # Cloud SQL instance connection name
+            "asyncpg",
+            user="my-user",
+            password="my-password",
+            db="my-db-name"
+            # ... additional database driver args
+        )
+        return conn
 
     # The Cloud SQL Python Connector can be used along with SQLAlchemy using the
     # 'async_creator' argument to 'create_async_engine'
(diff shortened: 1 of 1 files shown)


enocom (approved):
:ship:

PR #760, opened 2023-06-12
fix: improve timeout handling

### LINKED ISSUE #759: Connector `timeout` not being passed to drivers
### Bug Description

The `timeout` keyword arg for `Connector` and `connector.connect` is not fully working as intended and can be improved upon.  

It is being popped from the keyword args at the `Connector` level and used to wrap a timeout around the connection attempt. While this works, a better solution is to pass the `timeout` all the way down, directly to the database driver and have the database drivers handle the timeout natively and bubble up a more descriptive error.

### Example code (or command)

_No response_

### Stacktrace

_No response_

### Steps to reproduce?

Not needed
...

### Environment

Not needed

### Additional Details

_No response_

jackwotherspoon:
Allow database drivers to handle timeout instead of Connector itself.

Closes #759

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -227,7 +227,7 @@ class Connector:
             raise KeyError(f"Driver '{driver}' is not supported.")
 
         ip_type = kwargs.pop("ip_type", self._ip_type)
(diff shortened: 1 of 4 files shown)


jackwotherspoon:
Will look to add a test or two

enocom (approved):

PR #766, opened 2023-06-19
feat: add support for PSC connections

jackwotherspoon:
Add a new IP type `IPTypes.PSC` to support connections to Cloud SQL using [private service connect](https://cloud.google.com/vpc/docs/private-service-connect).

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,9 +245,15 @@ with pool.connect() as db_conn:
         print(row)
 ```
 
-### Specifying Public or Private IP
+### Specifying IP Address Type
 
-The Cloud SQL Connector for Python can be used to connect to Cloud SQL instances using both public and private IP addresses. To specify which IP address to use to connect, set the `ip_type` keyword argument Possible values are `IPTypes.PUBLIC` and `IPTypes.PRIVATE`.
+The Cloud SQL Python Connector can be used to connect to Cloud SQL instances
+using both public and private IP addresses, as well as
+[Private Service Connect][psc](PSC). To specify which IP address type to connect
+with, set the `ip_type` keyword argument when initializing a `Connector()` or when
+calling `connector.connect()`. Possible values are `IPTypes.PUBLIC` (default value),
+`IPTypes.PRIVATE`, and `IPTypes.PSC`.
+
 Example:
 
 ```python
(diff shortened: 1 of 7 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -238,6 +239,9 @@ async def connect_async(
         # attempt to make connection to Cloud SQL instance
         try:
             instance_data, ip_address = await instance.connect_info(ip_type)
+            # resolve DNS name into IP address for PSC
+            if ip_type.value == "PSC":
+                ip_address = socket.gethostbyname(ip_address)
```
enocom, line 244: Is there an async option for this?

enocom (approved):

google/cloud/sql/connector/connector.py
jackwotherspoon: @enocom there is no async version of `gethostbyname` however there is an async version of [`getaddrinfo`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.getaddrinfo) which can be used to retrieve the IP address. It is not as clean as you can see that it returns a list and then I am grabbing the first element of the list. It does seem to work though.

google/cloud/sql/connector/connector.py
```diff
@@ -238,6 +239,12 @@ async def connect_async(
         # attempt to make connection to Cloud SQL instance
         try:
             instance_data, ip_address = await instance.connect_info(ip_type)
+            # resolve DNS name into IP address for PSC
+            if ip_type.value == "PSC":
+                addr_info = await self._loop.getaddrinfo(
+                    ip_address, None, family=socket.AF_INET
+                )
+                ip_address = addr_info[0][4][0]
```
enocom, line 247: A comment here about these numbers would help. Also, do you see any reason to guard against these indexes being null?

enocom (approved):

google/cloud/sql/connector/connector.py
jackwotherspoon: I'll throw in an error message if the list is empty saying "Could not resolve PSC DNS name into IP addresss" or something

README.md
```diff
@@ -252,9 +252,15 @@ with Connector() as connector:
             print(row)
 ```
 
-### Specifying Public or Private IP
+### Specifying IP Address Type
+
+The Cloud SQL Python Connector can be used to connect to Cloud SQL instances
+using both public and private IP addresses, as well as
+[Private Service Connect][psc](PSC). To specify which IP address type to connect
```
hessjcg, line 259: It looks like maybe this isn't linked correctly.

jackwotherspoon: @hessjcg how so? Link works for me, I added a space between (PSC) so it looks visually better, otherwise I think the link is working.

PR #807, opened 2023-07-29
chore: add py.typed marker

danielroseman:
The library is fully typed, so add a marker to allow linters like mypy to pick this up as specified in PEP561.


jackwotherspoon (approved):
Thanks @danielroseman! LGTM much appreciated. 

Will leave the [PEP 561](https://peps.python.org/pep-0561/) link here for future reference.

PR #894, opened 2023-11-10
fix: refresh credentials for valid oauth2 token on each refresh

### LINKED ISSUE #892: Redundant instance refreshes (with IAM auth)
### Bug Description

Hi, after #890, number of refreshes dropped and it is much better. However, I noticed that a lot of refreshes still occur just before token expiration. It happens every 30 second (rate limiter in `Instance`) for those 4 minutes ([refresh-buffer](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/2a955322251a22e8c8b9d2494209b6db94b0d14d/google/cloud/sql/connector/refresh_utils.py#L36)) just before IAM token expiration.

The problem is that [here](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/2a955322251a22e8c8b9d2494209b6db94b0d14d/google/cloud/sql/connector/instance.py#L337), expiration is set to lower value of ephemeral cert or IAM credentials. But IAM credentials are refreshed only when they are invalid [here](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/2a955322251a22e8c8b9d2494209b6db94b0d14d/google/cloud/sql/connector/refresh_utils.py#L91) (or in get-ephemeral, but metadata are refreshed first).

Ephemeral cert expiration is usually 1 hour so let's say that we request instance refresh 30 minutes before IAM token expiration.
Expiration value sent to `_seconds_until_refresh()` computation will always be set to IAM token expiration because now it is just 30  minutes (and fewer) and its lower than ephemeral cert 1h. When it's less than 4 minutes before expiration, [this condition](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/refresh_utils.py#L228) applies and seconds-until refresh starts returning `0` - meaning, refresh is performed instantly (well, with 30 seconds rate limiter). And this repeats every 30 seconds until IAM token is refreshed.

I don't know why expiration is set to IAM token expiration, can't we remove [this block](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/2a955322251a22e8c8b9d2494209b6db94b0d14d/google/cloud/sql/connector/instance.py#L333-L337)? Ephemeral cert should be valid even when IAM token expired, shouldn't it?

I checked the golang implementation and it refreshes IAM token [always](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/8a1e5218b018ee1ef1134a6b3ee3f5671a106885/internal/cloudsql/refresh.go#L164), not only when it is invalid.

The point is, 4 minutes before IAM token expiration, next refresh is computed based on IAM token, but IAM token is not actually refreshed, only ephemeral cert is refreshed but it's expiration time is not considered when scheduling next refresh, so it keeps refreshing ephemeral cert but schedules next refresh based on IAM token again and again until IAM expires.

I would create a PR but I was not sure if my first suggestion (don't set expiration to IAM token) is OK, it worked just fine in my project, but maybe I missed something about tokens and their validity. I would probably go with the second suggestion (always refresh IAM) as it is already used in go and also [cloud-sql-proxy](https://github.com/GoogleCloudPlatform/cloud-sql-proxy) which I used as well and saw no problem with it (but Go refreshes token only in get-ephemeral, not in get-metadata).

(For sake of completeness of my ideas, the third option is to refresh IAM token in [_get_metadata](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/refresh_utils.py#L91) (and `_get_ephemeral`) when it is just about to expire, i.e. `credentials.expiry - utcnow() < _refresh_buffer`, if you don't want to refresh it always, but note that "always" means just once an hour (considering default expiration is 1h) so I think "refresh always" is simpler in this case and number of refreshes is same).

### Steps to reproduce?

You can use my code from #890. You have to find out when your IAM token expires and then run the script at least for that long. Refresh logs should appear every 30 seconds for 4 minutes before IAM token expiration.

```py
import logging
import time

import google.cloud.sql.connector
import sqlalchemy

instance_logger = logging.getLogger(name='google.cloud.sql.connector.instance')
instance_logger.setLevel(logging.DEBUG)

logging.basicConfig(format='%(asctime)s [%(levelname)s]: %(message)s')
logging.debug('test start')

connector = google.cloud.sql.connector.Connector()
def getconn():
    return connector.connect(
        "<project>:<region>:<db-name>",
        "pg8000",
        user="<username>",
        db="<db-name>",
        enable_iam_auth=True,
        ip_type=google.cloud.sql.connector.IPTypes.PRIVATE,
    )

pool = sqlalchemy.create_engine("postgresql+pg8000://", creator=getconn)

while True:
    with pool.connect() as db_conn:
        db_conn.execute(sqlalchemy.text("SELECT 1"))
    time.sleep(5)
```

### Environment

1. OS type and version: Linux, but this should not matter :)
2. Python version: 3.10
3. Cloud SQL Python Connector version:  current `main`, with #890 merged.

jackwotherspoon:
Previously, the credentials object which is used for auto IAM authentication was only refreshed when it became invalid (expired). This is buggy as the credentials expiration is sometimes used to calculate when to schedule the next refresh (when credentials expiration is sooner than cert expiration). If we do not refresh the credentials object as part of the refresh operation than we get constant refreshes scheduled in the final few minutes before the credentials' OAuth2 token expires.  

Fixes #892

### DIFF
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -288,7 +288,7 @@ class Instance:
 
             logger.debug(f"['{self._instance_connection_string}']: Creating context")
 
-            # refresh credentials for fresh Oauth2 token
+            # refresh credentials for fresh OAuth2 token
             request = google.auth.transport.requests.Request()
             self._credentials.refresh(request)
 
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -335,9 +335,10 @@ class Instance:
             )
             expiration = x509.not_valid_after
 
+            # for IAM authentication OAuth2 token is embedded in cert so it
+            # must still be valid for successful connection
             if self._enable_iam_auth:
-                if self._credentials is not None:
-                    token_expiration: datetime.datetime = self._credentials.expiry
+                token_expiration: datetime.datetime = self._credentials.expiry
                 if expiration > token_expiration:
                     expiration = token_expiration
 
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -33,6 +33,7 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.x509 import load_pem_x509_certificate
 
 from google.auth.credentials import Credentials
+import google.auth.transport.requests
 from google.cloud.sql.connector.exceptions import (
     AutoIAMAuthNotSupported,
     CloudSQLIPTypeError,
@@ -193,7 +194,7 @@ class Instance:
             self.__client_session = aiohttp.ClientSession(headers=headers)
         return self.__client_session
 
-    _credentials: Optional[Credentials] = None
+    _credentials: Credentials
     _keys: asyncio.Future
 
     _instance_connection_string: str
@@ -286,6 +287,10 @@ class Instance:
             priv_key, pub_key = await self._keys
 
             logger.debug(f"['{self._instance_connection_string}']: Creating context")
+
+            # refresh credentials for fresh Oauth2 token
+            request = google.auth.transport.requests.Request()
+            self._credentials.refresh(request)
 
             metadata_task = self._loop.create_task(
                 _get_metadata(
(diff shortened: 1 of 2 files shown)


enocom (approved):
This looks good to me.

jackwotherspoon:
Closing as #898 is a better fix

PR #904, opened 2023-11-16
chore: make driver versions more flexible

jackwotherspoon:
Driver packages like `asyncpg` have corresponding stub libraries such as [asyncpg-stubs](https://pypi.org/project/asyncpg-stubs/) that are aggressive on forcing latest version of driver for typing. Driver versions in the Connector may be slightly out of sync as we wait for next release at our monthly cadence. Being more lenient in driver version and not forcing the version to be explicit will allow users to not run into version conflicts between connector `asyncpg` driver version and stub library `asyncpg` version.

cc @kurtisvg

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,10 +74,10 @@ setup(
     namespace_packages=namespaces,
     install_requires=core_dependencies,
     extras_require={
-        "pymysql": ["PyMySQL==1.1.0"],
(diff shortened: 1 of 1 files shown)


jackwotherspoon:
Just need to test this but based on my research there is no issue using `>=` in extras.

enocom (approved):

PR #906, opened 2023-11-17
feat: introduce compatibility with native namespace packages

### LINKED ISSUE #876: Remove `declare_namespace` calls in package
Need to move off of using `pkg_resources.declare_namespace()` for package structure as it is deprecated.

```
DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google.cloud.sql')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
```

jackwotherspoon:
Fixes #876 

Ref: https://github.com/googleapis/google-auth-library-python/pull/1205


enocom (approved):
LGTM. Let's have a Python reviewable take a look and then we can merge this.

jackwotherspoon:
@parthea mind reviewing this PR as it is similar to those that you have done for the `googleapis` libraries? 😄

setup.py
```diff
@@ -35,20 +23,32 @@
...
 
+package_root = os.path.abspath(os.path.dirname(__file__))
+
+readme_filename = os.path.join(package_root, "README.md")
+with io.open(readme_filename, encoding="utf-8") as readme_file:
+    readme = readme_file.read()
+
+version = {}
+with open(os.path.join(package_root, "google/cloud/sql/connector/version.py")) as fp:
+    exec(fp.read(), version)
```
parthea, line 42: We will be removing `exec` from all libraries. This has already been added to `gapic-generator-python` and will be rolled out after the freeze. Please feel free to make the same change here or in a separate PR.

https://github.com/googleapis/gapic-generator-python/blob/2ad9be28500972777cec69097be6581eaefb89e4/gapic/templates/setup.py.j2#L22-L25

tests/unit/test_packaging.py
```diff
@@ -0,0 +1,31 @@
...
+
+import os
+import pathlib
+import subprocess
+import sys
+
+
+def test_namespace_package_compat(tmp_path: pathlib.PosixPath) -> None:
+    """
+    The ``google`` namespace package should not be masked
```
parthea, line 23: Please consider also adding checks for `google.cloud` and `google.cloud.sql`

parthea (approved):
LGTM. Please add the `do not merge` label to the release PR so this doesn't get merged during the freeze. Also, consider publishing a pre-release version with this change first to allow for additional compatibility testing.

enocom:
@parthea Merging this PR doesn't result in a release. Is there any reason why we shouldn't just merge this?

parthea:
@enocom I've blocked the PRs from being merged for python repositories in `googleapis` but this isn't a hard requirement. Please feel free to merge it. Please file a bug for the suggestion in https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/906#discussion_r1399356173 if it won't be addressed in this PR

setup.py
jackwotherspoon: Created issue for doing this in follow-up PR https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/issues/909

PR #923, opened 2023-12-01
WIP: asyncpg socket callback

### LINKED ISSUE #792: Create `asyncpg` connections using pre-configured SSLSocket
Need to update Python Connector to allow `asyncpg` connections through the use of a pre-configured SSLSocket or asyncio equivalent in order to unblock future IAM checks requiring a socket read/write prior to connecting to Cloud SQL.

Similar to [current pymysql implementation](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/pymysql.py#L53) but through the use of asyncio's `loop.create_connection()`

This will require an up-stream patch to the asyncpg library.

jackwotherspoon:
Add ability to create `asyncpg` connections with configured socket through `socket_callback` function.

Closes #792

### DIFF
--- a/google/cloud/sql/connector/asyncpg.py
+++ b/google/cloud/sql/connector/asyncpg.py
@@ -60,8 +60,8 @@ async def connect(
     async def async_sock_func() -> socket.socket:
         # TODO: switch to asyncio.to_thread once Python 3.8 support is dropped
         loop = asyncio.get_running_loop()
-        ctx = contextvars.copy_context()
-        func_call = functools.partial(ctx.run, sock_func, ip_address)
+        ctx_vars = contextvars.copy_context()
+        func_call = functools.partial(ctx_vars.run, sock_func, ip_address)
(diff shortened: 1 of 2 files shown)


google/cloud/sql/connector/asyncpg.py
```diff
@@ -48,6 +59,10 @@ async def connect(
         raise ImportError(
             'Unable to import module "asyncpg." Please install and try again.'
         )
+
+    async def async_sock_func() -> ssl.SSLSocket:
+        return await asyncio.to_thread(sock_func, ip_address, ctx)
```
jackwotherspoon, line 64: `asyncio.to_thread` was added in Python 3.9. Will have to implement this on my own to support Python 3.8.

jackwotherspoon:
Closing as socket callback is not the right approach, passing a connector factory is probably more favourable.

PR #942, opened 2023-12-21
chore: use connector factory for asyncpg connections

enocom:


enocom:
I think we can follow [the approach here](https://github.com/MagicStack/asyncpg/blob/master/asyncpg/connect_utils.py#L741) to gain access to the socket, do a read and write, close the transport, and then create a new connection that uses the database protocol.

enocom:
Depends on https://github.com/enocom/asyncpg/commit/2d3bbb8e90078fdae7fdb925e425d521b376f583 (which would have to be a PR into asyncpg)

PR #905, opened 2023-11-16
feat: add support for Python 3.12

jackwotherspoon:
Add that Python Connector officially supports [Python 3.12](https://www.python.org/downloads/release/python-3120/)

Closes #900

PR #965, opened 2024-01-11
chore: update to latest python in lint

enocom:


jackwotherspoon:
I've added this to #905 and will fix the latest lint problems as part of that PR

PR #983, opened 2024-01-25
chore: update datetime objects to be datetime aware

### LINKED ISSUE #982: Update datetimes to aware and remove naive datetimes
```
refresh_utils.py:214: CryptographyDeprecationWarning: Properties that return a naïve datetime object have been deprecated. Please switch to not_valid_after_utc.
```

jackwotherspoon:
Overall, datetime aware datetimes `datetime.now(timezone)` are recommended over datetime naive datetimes `datetime.utcnow()` etc and Python is beginning to slowly deprecate datetime naive as it leads to ambiguity overall. Switching over our repo to datetime aware.

Context: https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated

As well in cryptography library (v42.0.0) there is a new `not_valid_after_utc` datetime attribute that is datetime aware and is recommended over `not_valid_after` which is naive. A deprecation warning will be removed with this update.

Need to update `google.auth` credentials expiry to be timezone aware as [auth library strips the timezone info](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/_helpers.py#L93-L99) in order to be backwards compatible with python 2. 

Fixes #982

### DIFF
--- a/google/cloud/sql/connector/refresh_utils.py
+++ b/google/cloud/sql/connector/refresh_utils.py
@@ -196,7 +196,8 @@ async def _get_ephemeral(
     if enable_iam_auth:
         token_expiration: datetime.datetime = login_creds.expiry
         # google.auth library strips timezone info for backwards compatibality
-        # Ref: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/_helpers.py#L93-L99
+        # reasons with Python 2. Add it back to allow timezone aware datetimes.
+        # Ref: https://github.com/googleapis/google-auth-library-python/blob/49a5ff7411a2ae4d32a7d11700f9f961c55406a9/google/auth/_helpers.py#L93-L99
         token_expiration = token_expiration.replace(tzinfo=datetime.timezone.utc)
         if expiration > token_expiration:
             expiration = token_expiration
--- a/google/cloud/sql/connector/refresh_utils.py
+++ b/google/cloud/sql/connector/refresh_utils.py
@@ -211,7 +211,7 @@ async def _get_ephemeral(
 
     # decode cert to read expiration
     x509 = load_pem_x509_certificate(ephemeral_cert.encode("UTF-8"), default_backend())
-    expiration = x509.not_valid_after
+    expiration = x509.not_valid_after_utc
     # for IAM authentication OAuth2 token is embedded in cert so it
     # must still be valid for successful connection
     if enable_iam_auth:
@@ -234,7 +234,9 @@ def _seconds_until_refresh(
     :returns: Time in seconds to wait before performing next refresh.
     """
 
-    duration = int((expiration - datetime.datetime.utcnow()).total_seconds())
+    duration = int(
+        (expiration - datetime.datetime.now(datetime.timezone.utc)).total_seconds()
+    )
 
     # if certificate duration is less than 1 hour
     if duration < 3600:
@@ -251,7 +253,7 @@ async def _is_valid(task: asyncio.Task) -> bool:
     try:
         metadata = await task
         # only valid if now is before the cert expires
-        if datetime.datetime.utcnow() < metadata.expiration:
+        if datetime.datetime.now(datetime.timezone.utc) < metadata.expiration:
             return True
     except Exception:
         # supress any errors from task
--- a/google/cloud/sql/connector/refresh_utils.py
+++ b/google/cloud/sql/connector/refresh_utils.py
@@ -202,6 +202,13 @@ async def _get_ephemeral(
     # decode cert to read expiration
     x509 = load_pem_x509_certificate(ephemeral_cert.encode("UTF-8"), default_backend())
     expiration = x509.not_valid_after_utc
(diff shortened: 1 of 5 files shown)


jackwotherspoon:
Looks like google auth token expiry is not datetime aware, may need to wait till auth library makes the switch

google/cloud/sql/connector/refresh_utils.py
```diff
@@ -213,7 +216,9 @@ def _seconds_until_refresh(
     :returns: Time in seconds to wait before performing next refresh.
     """
 
-    duration = int((expiration - datetime.datetime.utcnow()).total_seconds())
+    duration = int(
+        (expiration - datetime.datetime.now(datetime.timezone.utc)).total_seconds()
```
enocom, line 220: What is the difference between:

```
 datetime.datetime.utcnow()
```

and

```
datetime.datetime.now(datetime.timezone.utc)
```

google/cloud/sql/connector/refresh_utils.py
```diff
@@ -190,11 +190,14 @@ async def _get_ephemeral(
 
     # decode cert to read expiration
     x509 = load_pem_x509_certificate(ephemeral_cert.encode("UTF-8"), default_backend())
-    expiration = x509.not_valid_after
+    expiration = x509.not_valid_after_utc
     # for IAM authentication OAuth2 token is embedded in cert so it
     # must still be valid for successful connection
     if enable_iam_auth:
         token_expiration: datetime.datetime = login_creds.expiry
+        # google.auth library strips timezone info for backwards compatibality
```
enocom, line 198: And so we're going to re-add the timezone (which is UTC)?

When the auth library strips the timezone, do they assume UTC afterwords?

Also nit: this GitHub reference will become stale as soon as those lines change. Shall we link to a SHA?

tests/unit/mocks.py
```diff
@@ -62,7 +64,11 @@ def expired(self) -> bool:
         """
         if self.expiry is None:
             return False
-        return False if self.expiry > datetime.datetime.utcnow() else True
+        return (
```
enocom, line 67: Could this be better expressed as:

```
if self.expiry > datetime.datetime.now(datetime.timezone.utc):
  return False

return True
```

?

tests/unit/mocks.py
```diff
@@ -108,11 +114,15 @@ def __init__(
 
 
 async def instance_metadata_success(*args: Any, **kwargs: Any) -> MockMetadata:
-    return MockMetadata(datetime.datetime.utcnow() + datetime.timedelta(minutes=10))
+    return MockMetadata(
+        datetime.datetime.now(datetime.timezone.utc) + datetime.timedelta(minutes=10)
```
enocom, line 118: What is the significance of 10? Ditto below.

tests/unit/test_refresh_utils.py
```diff
@@ -226,7 +225,11 @@ def test_seconds_until_refresh_over_1_hour() -> None:
     # using pytest.approx since sometimes can be off by a second
     assert (
         pytest.approx(
-            _seconds_until_refresh(datetime.utcnow() + timedelta(minutes=62)), 1
+            _seconds_until_refresh(
+                datetime.datetime.now(datetime.timezone.utc)
+                + datetime.timedelta(minutes=62)
```
enocom, line 230: Same here, but with 62. What is 62?

google/cloud/sql/connector/refresh_utils.py
jackwotherspoon: Yup we are going to re-add the timezone information to the datetime, its already in UTC based on the linked code. So adding it back should have no affect but I'll do a bit more research to be 100% sure that stripping and re-adding is okay.

tests/unit/mocks.py
jackwotherspoon: yup, had it all one line but looks like formatter edited it

google/cloud/sql/connector/refresh_utils.py
enocom: Let's add that as a comment to make it crystal clear (or the commit message).

google/cloud/sql/connector/refresh_utils.py
enocom: Sounds good. Let's just make the rationale clear in the code, so future maintainers don't trip over this.

tests/unit/test_refresh_utils.py
jackwotherspoon: this is explained in the docstring of the test... testing our scheduling logic for cert expiring in over an hour
```python
"""
Test _seconds_until_refresh returns proper time in seconds.
    If expiration is over 1 hour, should return duration/2.
"""
```

tests/unit/mocks.py
jackwotherspoon: No real significance, just a trivial time in the future for tests to pass.

PR #986, opened 2024-01-26
feat: add support for a custom user agent

enocom:


google/cloud/sql/connector/connector.py
```diff
@@ -91,6 +91,7 @@ def __init__(
         loop: Optional[asyncio.AbstractEventLoop] = None,
         quota_project: Optional[str] = None,
         sqladmin_api_endpoint: str = "https://sqladmin.googleapis.com",
+        custom_user_agent: Optional[str] = None,
```
jackwotherspoon, line 94: nit: the fact that the user is configuring makes it "custom", redundant to have it in the var name
```suggestion
        user_agent: Optional[str] = None,
```

google/cloud/sql/connector/connector.py
```diff
@@ -115,6 +116,7 @@ def __init__(
         self._quota_project = quota_project
         self._sqladmin_api_endpoint = sqladmin_api_endpoint
         self._credentials = credentials
+        self._custom_user_agent = custom_user_agent
```
jackwotherspoon, line 119: I am fine leaving this as `self._custom_user_agent` for clarity since it is an internal

google/cloud/sql/connector/connector.py
```diff
@@ -211,6 +213,7 @@ async def connect_async(
                 enable_iam_auth,
                 self._quota_project,
                 self._sqladmin_api_endpoint,
+                custom_user_agent=self._custom_user_agent,
```
jackwotherspoon, line 216: nit: same as above, lets make this `user_agent`

google/cloud/sql/connector/instance.py
```diff
@@ -134,6 +134,13 @@ def get_preferred_ip(self, ip_type: IPTypes) -> str:
         )
 
 
+def _format_user_agent(version: str, driver: str, custom: Optional[str]) -> str:
+    agent = f"{APPLICATION_NAME}/{version}+{driver}"
+    if custom:
+        agent = " ".join([agent, custom])
```
jackwotherspoon, line 140: is it normal to have a whitespace in a user agent? Would this make this harder to parse for internal tools?

jackwotherspoon (changes requested):

google/cloud/sql/connector/instance.py
enocom: Yes. It's a standard. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent. We do the same in the Proxy. https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/cmd/root.go#L656

google/cloud/sql/connector/connector.py
enocom: Fixed.

google/cloud/sql/connector/connector.py
enocom: I changed it to user agent -- see how that looks.

jackwotherspoon (approved):

Issue #873, opened 2023-10-24
Use one client session per connector


enocom:
Currently, we use one client session per instance. Instead, we should use one client session per connector.

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/32c64817aa1a3e5c3a86409535d57568a5782852/google/cloud/sql/connector/instance.py#L184-L194

PR #995, opened 2024-01-31
chore: refactor credentials to be initialized in Connector

jackwotherspoon:
Refactor the `credentials` from being initialized at the `Instance` level to the `Connector` level.

 Improvement already made as a new credentials object was being used for each `Instance`, now done once and shared across all instances.

Part of the work to move to using one client session per Connector #873

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -21,7 +21,11 @@ import logging
 import socket
 from threading import Thread
 from types import TracebackType
-from typing import Any, Dict, Optional, Type, TYPE_CHECKING
+from typing import Any, Dict, Optional, Type
+
+import google.auth
+from google.auth.credentials import Credentials
+from google.auth.credentials import with_scopes_if_required
 
 import google.cloud.sql.connector.asyncpg as asyncpg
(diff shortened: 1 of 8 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -109,13 +110,24 @@ def __init__(
             )
         self._instances: Dict[str, Instance] = {}
 
+        # initialize credentials
+        scopes = ["https://www.googleapis.com/auth/sqlservice.admin"]
+        if credentials:
+            if not isinstance(credentials, Credentials):
```
enocom, line 116: Do we do this because there are other credential-like objects out there?

enocom (approved):

google/cloud/sql/connector/connector.py
jackwotherspoon: This is the base class for credentials, there are several child types that can be supported so we just are checking that its atleast the base class

enocom: Adding that as a comment might be nice.

jackwotherspoon: done

PR #996, opened 2024-01-31
chore: use better fake for credentials

### LINKED ISSUE #901: Create better mock for `Credentials` in tests
Currently the mocked credentials fixture does not behave nicely when it is refreshed. This bug is to improve the `fake_credentials` text fixture so it can be properly refreshed in tests.

Re-enable `test_downscope_credentials_service_account` test in `test_refresh_utils.py`

jackwotherspoon:
Previous fake credentials object was copy/pasted from random code somewhere.

Updating fake credentials to have more control, such as setting expiry etc.

Closes #901

### DIFF
--- a/tests/unit/test_refresh_utils.py
+++ b/tests/unit/test_refresh_utils.py
@@ -267,22 +267,6 @@ async def test_is_valid_with_expired_metadata() -> None:
     assert not await _is_valid(task)
 
 
-# TODO: https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/issues/901
-# def test_downscope_credentials_service_account(fake_credentials: Credentials) -> None:
-#     """
-#     Test _downscope_credentials with google.oauth2.service_account.Credentials
-#     which mimics an authenticated service account.
-#     """
-#     # override actual refresh URI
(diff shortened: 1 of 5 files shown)


tests/unit/test_refresh_utils.py
```diff
@@ -287,22 +267,6 @@ async def test_is_valid_with_expired_metadata() -> None:
...
-#     # override actual refresh URI
-#     setattr(fake_credentials, "with_scopes", google.auth.credentials.Credentials(scopes=["https://www.googleapis.com/auth/sqlservice.login"]))
-#     credentials = _downscope_credentials(fake_credentials)
-#     # verify default credential scopes have not been altered
-#     assert fake_credentials.scopes == SCOPES
-#     # verify downscoped credentials have new scope
-#     assert credentials.scopes == ["https://www.googleapis.com/auth/sqlservice.login"]
-#     assert credentials != fake_credentials
-
-
```
jackwotherspoon, line 305: there really is not much of a point to this test, this test is what our actual iam authn integration tests do and test so no point of "faking it"

enocom (approved):

PR #997, opened 2024-02-01
chore: remove unreachable code

jackwotherspoon:
When we added support for domain-scoped instance connection names #937 we added regex for validating the instance connection name and parsing it into `project`, `region` and `instance` strings.

This string parsing guarantees that if completed the three arguments will be strings so it is unnecessary and useless to check the types of these variables later in the flow, if the execution has made it to the API calls, the variables must be strings.


enocom (approved):
Nice cleanup.

PR #1016, opened 2024-02-16
fix: update create_async_connector args

kurtisvg:
Update the `create_async_connector` args so they are consistent with the Connector.


jackwotherspoon:
@kurtisvg looks like the lint job is complaining

jackwotherspoon (approved):

PR #1029, opened 2024-03-04
feat: support `ip_type` as str

jackwotherspoon:
Adding support for `ip_type` as `str` type with options as `"public"`, `"private"` or `"psc"`.

If users want to configure a Connector for a specific IP type they currently need to import `IPTypes` first and then set `ip_type=IPTypes.PUBLIC` in the Connector config or during their call to connect.

```python
from google.cloud.sql.connector import IPTypes
```

This is an extra step that is unnecessary and causes friction to users, we should support users passing in the IPTypes enum or string value. This will allow users to not have to import the additional class.

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -133,18 +133,9 @@ class Connector:
         self._quota_project = quota_project
         self._sqladmin_api_endpoint = sqladmin_api_endpoint
         self._user_agent = user_agent
-        # if ip_type is str, convert to enum
+        # if ip_type is str, convert to IPTypes enum
         if isinstance(ip_type, str):
-            if ip_type.lower() == "public":
-                ip_type = IPTypes.PUBLIC
-            elif ip_type.lower() == "private":
-                ip_type = IPTypes.PRIVATE
-            elif ip_type.lower() == "psc":
-                ip_type = IPTypes.PSC
-            else:
-                raise ValueError(
-                    f"Incorrect value for ip_type, got '{ip_type}'. Want one of: 'public', 'private' or 'psc'."
-                )
+            ip_type = IPTypes._get_ip_type_from_str(ip_type)
         self._ip_type = ip_type
 
     def connect(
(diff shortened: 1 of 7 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -130,10 +130,22 @@ def __init__(
...
+                ip_type = IPTypes.PUBLIC
+            elif ip_type.lower() == "private":
+                ip_type = IPTypes.PRIVATE
+            elif ip_type.lower() == "psc":
+                ip_type = IPTypes.PSC
+            else:
+                raise ValueError(
+                    f"Incorrect value for ip_type, got '{ip_type}'. Want one of: 'public', 'private' or 'psc'."
+                )
+        self._ip_type = ip_type
```
jackwotherspoon, line 148: Technically this code could be removed and we could allow the conversion to purely happen once during the `connect` call. 

However, this would cause the case where users configure the ip type in the Connector() to be converted repeatedly on every single call to connect. For this reason I thought it may be worth converting in the Connector.__init__ so this case is converted once and then skips the check in the `connect` for reduced latency on connect.

google/cloud/sql/connector/connector.py
```diff
@@ -130,10 +130,22 @@ def __init__(
         # set default params for connections
         self._timeout = timeout
         self._enable_iam_auth = enable_iam_auth
-        self._ip_type = ip_type
         self._quota_project = quota_project
         self._sqladmin_api_endpoint = sqladmin_api_endpoint
         self._user_agent = user_agent
+        # if ip_type is str, convert to enum
+        if isinstance(ip_type, str):
```
enocom, line 137: nit: this and the code below could be wrapped up into a private function to avoid duplication and help long term maintenance.

enocom (approved):

google/cloud/sql/connector/connector.py
jackwotherspoon: Added the code as a private static method on the `IPTypes` class, this way it is marked as a utility function and tightly coupled with the class.

enocom (approved):

PR #1031, opened 2024-03-07
refactor: update IPTypes

jackwotherspoon:
Aligning approach as close as possible to https://github.com/GoogleCloudPlatform/alloydb-python-connector/pull/267

Using built-in `__missing__` to better raise custom error message: https://stackoverflow.com/questions/68899057/can-one-add-a-custom-error-in-enum-to-show-valid-values

### DIFF
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -64,20 +64,19 @@ class IPTypes(Enum):
     PRIVATE: str = "PRIVATE"
     PSC: str = "PSC"
 
-    @staticmethod
-    def _get_ip_type_from_str(ip_type_str: str) -> IPTypes:
-        """Utility method to convert IP type from a str into IPTypes."""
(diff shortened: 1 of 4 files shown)


google/cloud/sql/connector/instance.py
```diff
@@ -64,20 +64,19 @@ class IPTypes(Enum):
...
-            )
-        return ip_type
+    @classmethod
+    def _missing_(cls, value: object) -> None:
+        raise ValueError(
+            f"Incorrect value for ip_type, got '{value}'. Want one of: "
+            f"{', '.join([repr(m.value) for m in cls])}, 'PUBLIC'."
+        )
+
+    @classmethod
```
jackwotherspoon, line 74: updated this from `@staticmethod` to `@classmethod` since we were returning `IPTypes` which is the class itself hence by definition a class method

enocom (approved):

PR #1052, opened 2024-03-26
chore: verify custom user agent is type str

jackwotherspoon:
If custom user agent is not of type `str`, do not use it and use default user agent.


ttosta-google (approved):
LGTM

PR #1079, opened 2024-05-09
refactor: move SSL/TLS context creation to ConnectionInfo

jackwotherspoon:
This refactor moves all SSL/TLS configuration into its own method `create_ssl_context()` on the `ConnectionInfo` class.

This way `ConnectionInfo()` prepares all the info required to connect and then `create_ssl_context` will use the info to establish the SSL/TLS connection when called from `Connector.connect` at the time of connection.

This refactor also improves the testing infra to be more in line with AlloyDB Python Connector.

Based on Go: https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/pull/761

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -266,7 +266,10 @@ class Connector:
                 )
         else:
             cache = RefreshAheadCache(
-                instance_connection_string, self._client, self._keys, enable_iam_auth
+                instance_connection_string,
+                self._client,
+                self._keys,
+                enable_iam_auth,
             )
             self._cache[instance_connection_string] = cache
 
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -269,6 +269,7 @@ class Connector:
                 instance_connection_string,
                 self._client,
                 self._keys,
+                enable_iam_auth
             )
             self._cache[instance_connection_string] = cache
 
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -266,10 +266,7 @@ class Connector:
                 )
         else:
             cache = RefreshAheadCache(
-                instance_connection_string,
-                self._client,
-                self._keys,
-                enable_iam_auth
+                instance_connection_string, self._client, self._keys, enable_iam_auth
             )
             self._cache[instance_connection_string] = cache
 
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -269,7 +269,6 @@ class Connector:
                 instance_connection_string,
                 self._client,
                 self._keys,
-                enable_iam_auth,
             )
             self._cache[instance_connection_string] = cache
 
@@ -330,10 +329,17 @@ class Connector:
 
             # async drivers are unblocking and can be awaited directly
             if driver in ASYNC_DRIVERS:
-                return await connector(ip_address, instance_data.context, **kwargs)
+                return await connector(
+                    ip_address,
+                    instance_data.create_ssl_context(enable_iam_auth),
+                    **kwargs,
+                )
             # synchronous drivers are blocking and run using executor
             connect_partial = partial(
-                connector, ip_address, instance_data.context, **kwargs
+                connector,
+                ip_address,
+                instance_data.create_ssl_context(enable_iam_auth),
+                **kwargs,
             )
             return await self._loop.run_in_executor(None, connect_partial)
 
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -17,6 +17,7 @@ limitations under the License.
 from __future__ import annotations
 
 import asyncio
+from dataclasses import dataclass
 from enum import Enum
 import logging
 import re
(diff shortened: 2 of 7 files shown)


google/cloud/sql/connector/instance.py
```diff
@@ -303,6 +302,11 @@ async def _refresh_task(self: RefreshAheadCache, delay: int) -> ConnectionInfo:
                     await asyncio.sleep(delay)
                 refresh_task = asyncio.create_task(self._perform_refresh())
                 refresh_data = await refresh_task
+                # check that refresh is valid
+                if not await _is_valid(refresh_task):
+                    raise RefreshNotValidError(
+                        f"['{self._instance_connection_string}']: Invalid refresh operation. Certficate appears to be expired."
+                    )
```
jackwotherspoon, line 309: we do this in AlloyDB: https://github.com/GoogleCloudPlatform/alloydb-python-connector/blob/bb5d107b801e713b4e94c8dc644e6950717704cb/google/cloud/alloydb/connector/instance.py#L196-L199

This check fixed a test that was passing due to bad legacy test infra.

tests/unit/mocks.py
```diff
@@ -85,36 +84,6 @@ def valid(self) -> bool:
...
-async def instance_metadata_expired(*args: Any, **kwargs: Any) -> MockMetadata:
-    return MockMetadata(
-        datetime.datetime.now(datetime.timezone.utc) - datetime.timedelta(minutes=10)
-    )
-
-
-async def instance_metadata_error(*args: Any, **kwargs: Any) -> None:
-    raise BadRefresh("something went wrong...")
-
-
```
jackwotherspoon, line 117: this legacy mock was hiding a bad test, removed it and now our tests properly rely on our internals.

google/cloud/sql/connector/connector.py
```diff
@@ -330,10 +330,17 @@ async def connect_async(
 
             # async drivers are unblocking and can be awaited directly
             if driver in ASYNC_DRIVERS:
-                return await connector(ip_address, instance_data.context, **kwargs)
+                return await connector(
+                    ip_address,
+                    instance_data.create_ssl_context(enable_iam_auth),
```
jackwotherspoon, line 335: i guess the downside of this is now we are creating an SSL/TLS context per connection vs in the past where we were creating it once per `ConnectionInfo` class...

enocom: Is there a significant cost associated with creating an SSL context per connection? What would a Python developer expect?

google/cloud/sql/connector/instance.py
```diff
@@ -79,33 +81,31 @@ def _from_str(cls, ip_type_str: str) -> IPTypes:
         return cls(ip_type_str.upper())
 
 
+@dataclass
```
enocom, line 84: :heart:

google/cloud/sql/connector/instance.py
```diff
@@ -119,18 +119,18 @@ def __init__(
...
         # and ephemeral cert are loaded into the SSLcontext. The values
         # need to be written to files in order to be loaded by the SSLContext
         with TemporaryDirectory() as tmpdir:
             ca_filename, cert_filename, key_filename = write_to_file(
-                tmpdir, server_ca_cert, ephemeral_cert, private_key
+                tmpdir, self.server_ca_cert, self.client_cert, self.private_key
             )
-            self.context.load_cert_chain(cert_filename, keyfile=key_filename)
-            self.context.load_verify_locations(cafile=ca_filename)
+            context.load_cert_chain(cert_filename, keyfile=key_filename)
```
enocom, line 131: Are there other ways to load this data? If each connection has to read from the disk, this will result in worse performance.

enocom: Maybe we could cache the SSL context so that we only load it once?

jackwotherspoon: I added caching the context for creating it only once per `ConnectionInfo`

google/cloud/sql/connector/instance.py
```diff
@@ -79,33 +81,34 @@ def _from_str(cls, ip_type_str: str) -> IPTypes:
...
-        server_ca_cert: str,
-        expiration: datetime.datetime,
-        enable_iam_auth: bool,
-    ) -> None:
-        self.ip_addrs = ip_addrs
-        self.database_version = database_version
-        self.context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+    def create_ssl_context(self, enable_iam_auth: bool = False) -> ssl.SSLContext:
+        """Constructs a SSL/TLS context for the given connection info."""
+        if self.context is not None:
```
enocom, line 99: Shall we add a comment to remind readers why we're caching the context? e.g., "Cache the SSL context to ensure we don't read from disk repeatedly when configuring a secure connection."

enocom (commented):
Do we want to test that the ConnectionInfo caches its ssl context?

This could be a simple verification of object identity being the same from one call to the next.

google/cloud/sql/connector/instance.py
jackwotherspoon: good call, will add one

enocom (approved):

PR #1088, opened 2024-05-23
chore: use rendered notices in README

jackwotherspoon:
Add use of rendered notices in README.

image

Also updating SQL Server verbiage slightly


hessjcg (approved):
LGTM

PR #1089, opened 2024-05-24
refactor: use google.auth TokenState for credentials validity

### LINKED ISSUE #1048: Replace`credentials.valid` check for `credentials.token_state`
New best practice for checking validity of credentials is no longer `credentials.valid` which is now deprecated (as of google-auth v2.24.0). Instead the google-auth folks recommend using [`credentials.token_state`](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/credentials.py#L100)

All instances of the following:
https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/4d5e2b3cc03ed7ceb4c5463fd373a90c00f871dd/google/cloud/sql/connector/client.py#L116-L118

Should be replaced to:
```python
from google.auth.credentials import TokenState

# ...
if not self._credentials.token_state == TokenState.FRESH:
    self._credentials.refresh(google.auth.transport.requests.Request())
```

jackwotherspoon:
In the google.auth package, it is now recommended to use `.token_state`
over `.valid` for checking validity of credentials.

Token state gives more information such as if token is stale, expired,
invalid etc.

`.valid` is now marked as deprecated ([code](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/credentials.py#L87C5-L97C59)):
```python
@property
def valid(self):
    """Checks the validity of the credentials.

    This is True if the credentials have a :attr:`token` and the token
    is not :attr:`expired`.

    .. deprecated:: v2.24.0
      Prefer checking :attr:`token_state` instead.
    """
    return self.token is not None and not self.expired
```

Closes #1048

### DIFF
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -26,6 +26,8 @@ from tempfile import TemporaryDirectory
 from typing import Any, Dict, Tuple, TYPE_CHECKING
 
 import aiohttp
+from google.auth.credentials import TokenState
+from google.auth.transport import requests
 
 from google.cloud.sql.connector.client import CloudSQLClient
 from google.cloud.sql.connector.exceptions import AutoIAMAuthNotSupported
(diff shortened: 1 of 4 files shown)


google/cloud/sql/connector/instance.py
```diff
@@ -231,6 +233,10 @@ async def _perform_refresh(self) -> ConnectionInfo:
 
             logger.debug(f"['{self._instance_connection_string}']: Creating context")
 
+            # before making Cloud SQL Admin API calls, refresh creds
+            if not self._client._credentials.token_state == TokenState.FRESH:
+                self._client._credentials.refresh(requests.Request())
+
```
jackwotherspoon, line 239: moved checking of credentials up a level into the refresh operation and out of the individual API calls, this way we are checking once instead of twice.

enocom (approved):

Issue #805, opened 2023-07-26
Add support for lazy certificate refresh


enocom:
### Feature Description

In serverless environments, when the CPU is throttled, the Python Connector will sometimes fail to refresh the certificate.

This is a feature request to allow callers to opt-in to a lazy refresh strategy, where no background tasks run and the cert is refreshed as needed.

enocom:
Related to https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/992.

jackwotherspoon:
Closed via #1093

PR #1090, opened 2024-05-29
refactor: move ConnectionInfo to own file and add `get_connection_info`

jackwotherspoon:
Refactor revolving around `ConnectionInfo` in preparation for adding a
`LazyRefreshCache`.

Moving `ConnectionInfo` into its own file, `connection_info.py` as it is
to be shared by both `RefreshAheadCache` and `LazyRefreshCache`.

Adding a `get_connection_info` method to the `CloudSQLClient`.
This is the equivalent of the Go Connector's [refresher.ConnectInfo](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/4b68de3693e25642acd847d0c8ac393982d00c9b/internal/cloudsql/refresh.go#L274).

This will allow the lazy refresh to just check expiration and then call
`get_connection_info` and not need to duplicate code from refresh
ahead cache.

Related to #805

### DIFF
--- a/google/cloud/sql/connector/client.py
+++ b/google/cloud/sql/connector/client.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import datetime
 import logging
 from typing import Any, Dict, Optional, Tuple, TYPE_CHECKING
@@ -21,7 +22,11 @@ from typing import Any, Dict, Optional, Tuple, TYPE_CHECKING
 import aiohttp
 from cryptography.hazmat.backends import default_backend
 from cryptography.x509 import load_pem_x509_certificate
+from google.auth.credentials import TokenState
+from google.auth.transport import requests
 
+from google.cloud.sql.connector.connection_info import ConnectionInfo
+from google.cloud.sql.connector.exceptions import AutoIAMAuthNotSupported
 from google.cloud.sql.connector.refresh_utils import _downscope_credentials
 from google.cloud.sql.connector.version import __version__ as version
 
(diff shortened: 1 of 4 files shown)


hessjcg (approved):

google/cloud/sql/connector/client.py
```diff
@@ -212,6 +217,80 @@ async def _get_ephemeral(
...
+            )
+        )
+        try:
+            metadata = await metadata_task
+            # check if automatic IAM database authn is supported for database engine
+            if enable_iam_auth and not metadata["database_version"].startswith(
+                ("POSTGRES", "MYSQL")
+            ):
+                raise AutoIAMAuthNotSupported(
+                    f"'{metadata['database_version']}' does not support automatic IAM authentication. It is only supported with Cloud SQL Postgres or MySQL instances."
```
enocom, line 276: Can we wrap this at ~80-ish lines? Maybe like so:

``` python
f"'{metadata['database_version']}' does not support " +
  "automatic IAM authentication. It is only supported " +
  "with Cloud SQL Postgres or MySQL instances.
```

google/cloud/sql/connector/connection_info.py
```diff
@@ -0,0 +1,105 @@
...
+
+if TYPE_CHECKING:
+    import datetime
+
+    from google.cloud.sql.connector.instance import IPTypes
+
+logger = logging.getLogger(name=__name__)
+
+
+@dataclass
```
enocom, line 35: :clap: These are great.

enocom (approved):

google/cloud/sql/connector/client.py
jackwotherspoon: done

PR #1092, opened 2024-05-29
refactor: remove `ip_type` from connect_info

jackwotherspoon:
Remove `ip_type` from `RefreshAheadCache.connect_info()`.

The `connect_info` method should return the `ConnectionInfo`
for the given cache and that is all. 

Currently, it gets the connection info, then gets the ip address of 
the connection info and returns both. This is just not necessary as
the ConnectionInfo already contains the method to get the
preferred IP address so we should return the `ConnectionInfo` as
is and have the `Connector.connect` grab the preferred IP for a
more straightforward flow.

This matches the Go Connectors use of [`ConnectionInfo.Addr`](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/main/internal/cloudsql/instance.go#L200)

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -300,7 +300,8 @@ class Connector:
 
         # attempt to make connection to Cloud SQL instance
         try:
-            instance_data, ip_address = await cache.connect_info(ip_type)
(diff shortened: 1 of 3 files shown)


enocom (approved):
:clap:

PR #1093, opened 2024-05-30
feat: add support for lazy refresh strategy

jackwotherspoon:
Add `refresh_strategy` argument to `Connector()` that allows setting
the strategy to `"lazy"` to use a lazy refresh strategy.

When creating a Connector via `Connector(refresh_strategy="lazy")`,
the connection info and ephemeral certificate will be refreshed only
when the cache certificate has expired. No background tasks run
periodically with this option, making it ideal for use in serverless
environments such as Cloud Run, Cloud Functions, etc, where the
CPU may be throttled.

Usage example:

```python

with Connector(refresh_strategy="lazy") as connector:
    # ... use connector in lazy refresh mode
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,14 +185,15 @@ defaults for each connection to make, you can initialize a
 `Connector` object as follows:
 
 ```python
-from google.cloud.sql.connector import Connector, IPTypes
+from google.cloud.sql.connector import Connector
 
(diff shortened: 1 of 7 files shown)


enocom (approved):
:ship:

PR #1095, opened 2024-05-30
fix: use asyncio.Lock over Event

jackwotherspoon:
We are currently improperly using [`asyncio.Event` ](https://docs.python.org/3/library/asyncio-sync.html#event) in our code. Event is meant to be used with the `wait` coroutine to hold coroutine from running until the `set()` is called. We don't use wait anywhere in our code thus this is not having the intended effect. 

Much easier to just use [`asyncio.Lock`](https://docs.python.org/3/library/asyncio-sync.html#lock) which allows usage as a context manager to acquire/release the mutex lock.

### DIFF
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -129,7 +129,7 @@ class RefreshAheadCache:
             max_capacity=2,
             rate=1 / 30,
         )
-        self._refresh_in_progress = asyncio.locks.Event()
+        self._lock = asyncio.Lock()
         self._current: asyncio.Task = self._schedule_refresh(0)
         self._next: asyncio.Task = self._current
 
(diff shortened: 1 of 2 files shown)


enocom (commented):
We might need to carefully audit all the reads and writes to `self._current` and `self._next`.  I'd expect to see us holding the lock whenever we read or write from those values. WDYT?

jackwotherspoon:
Still have a couple more spots to acquire the lock, the tricky one is in the cache constructor as Python `__init__` functions are sync by default. Might have to add an async class method and use it to initialize the cache so that I can properly use the lock.

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/f8de0f11b1d48a920ce384713bb6ef2857309f7e/google/cloud/sql/connector/instance.py#L97-L99

jackwotherspoon:
Closing this as stale

PR #1108, opened 2024-06-03
feat: validate driver with database engine

### LINKED ISSUE #998: Validate driver matches database type
The behavior where someone passes a valid instance connection name for a Cloud SQL instance but for the wrong driver is slightly ambiguous currently. (ex. instance connection name for a Cloud SQL Postgres instance but with `pymysql` driver)

We should make an enum mapping drivers to database type and verify that the driver being used is correct before attempting a connection. This way we can throw a helpful error message to users pointing at the wrong instance.

We are already returning the database version as part of the metadata request so this should be pretty straightforward.

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/25a740a5db88088082820b6ae72f4b4a2a52dd2d/google/cloud/sql/connector/refresh_utils.py#L110-L114

jackwotherspoon:
Removing ambiguous behavior if a user configures a database
driver to connect to an incompatible database engine.

i.e. `pymysql` (MySQL driver) -> Cloud SQL for PostgreSQL database

This behavior would timeout trying to connect and then throw
an ambiguous timeout error.

The `Connector.connect` already know which database engine
the Cloud SQL database is prior to attempting to connect via the
driver. So we can validate if the driver is compatible and throw an
actionable error if not.

Closes #998

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -33,7 +33,6 @@ from google.cloud.sql.connector.client import CloudSQLClient
 from google.cloud.sql.connector.enums import DriverMapping
 from google.cloud.sql.connector.exceptions import ConnectorLoopError
 from google.cloud.sql.connector.exceptions import DnsNameResolutionError
-from google.cloud.sql.connector.exceptions import IncompatibleDriverError
 from google.cloud.sql.connector.instance import IPTypes
 from google.cloud.sql.connector.instance import RefreshAheadCache
 from google.cloud.sql.connector.instance import RefreshStrategy
@@ -335,13 +334,7 @@ class Connector:
         try:
             conn_info = await cache.connect_info()
             # validate driver matches intended database engine
-            mapping = DriverMapping[driver.upper()]
-            if not conn_info.database_version.startswith(mapping.value):
-                raise IncompatibleDriverError(
-                    f"Database driver '{driver}' is incompatible with database "
-                    f"version '{conn_info.database_version}'. Given driver can "
-                    f"only be used with Cloud SQL {mapping.value} databases."
-                )
+            DriverMapping.validate_engine(driver, conn_info.database_version)
             ip_address = conn_info.get_preferred_ip(ip_type)
             # resolve DNS name into IP address for PSC
             if ip_type.value == "PSC":
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -30,8 +30,10 @@ from google.auth.credentials import with_scopes_if_required
 
 import google.cloud.sql.connector.asyncpg as asyncpg
 from google.cloud.sql.connector.client import CloudSQLClient
+from google.cloud.sql.connector.enums import DriverMapping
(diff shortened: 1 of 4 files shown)


google/cloud/sql/connector/enums.py
```diff
@@ -0,0 +1,24 @@
...
+from enum import Enum
+
+
+class DriverMapping(Enum):
+    """Maps a given database driver to it's corresponding database engine."""
+
+    ASYNCPG = "POSTGRES"
+    PG8000 = "POSTGRES"
+    PYMYSQL = "MYSQL"
+    PYTDS = "SQLSERVER"
```
jackwotherspoon, line 24: I am going to have a follow-up refactor PR that moves the other shared enum types (`IPTypes`, `RefreshStrategy`) out of `instance.py` and into this file.

google/cloud/sql/connector/connector.py
```diff
@@ -321,6 +323,14 @@ async def connect_async(
         # attempt to make connection to Cloud SQL instance
         try:
             conn_info = await cache.connect_info()
+            # validate driver matches intended database engine
+            mapping = DriverMapping[driver.upper()]
```
enocom, line 327: What if we encapsulated more details and did something like this?

``` python
conn_info = await cache.connect_info()
if not DriverMapping.matching_engine(driver, conn_info.database_version):
    raise IncompatibleDriverError
```

where the DriverMapping method looks like:

``` python
class DriverMapping(Enum):
    def matching_engine(driver: str, engine: str) -> bool:
        return engine.startswith(DriverMapping[driver.upper()])
```

enocom (approved):

google/cloud/sql/connector/connector.py
```diff
@@ -321,6 +323,14 @@ async def connect_async(
...
+            if not conn_info.database_version.startswith(mapping.value):
+                raise IncompatibleDriverError(
+                    f"Database driver '{driver}' is incompatible with database "
```
hessjcg, line 330: This is a fatal configuration error. I think the message could be more clear. Something like:

"Fatal Error: the Cloud SQL Connector is configured with an incompatible database driver. Configured driver '{driver}' cannot connect to database instance '{instance_connection_string}' running version {conn_info.database_version}."

google/cloud/sql/connector/connector.py
jackwotherspoon: this would mean i no longer have access to the proper compatible database engine `mapping.value` in the error message without doing a separate call to `DriverMapping[driver.upper()].value` which is more work...

enocom: Just an idea -- feel free to handle it however you like. A good error message is most important.

google/cloud/sql/connector/connector.py
jackwotherspoon: @hessjcg  I don't normally see Python errors get prefixed with "Fatal Error:", its more the assumption that if an error is raised from the application it must be fatal.

I like the error message as is personally, concise and actionable. I don't have access to `instance_connection_string` from where the error is raised after the encapsulation refactor...

Issue #1107, opened 2024-05-31
RuntimeError: Timeout context manager should be used inside a task


colonelpanic8:
### Bug Description

This can occur when sqlalchemy attempts to create a new connection in the context of event loop that is not the event loop that cloud-sql-python-connector is using.

The connector has a member `_client`:

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/de2852f1e5333449f3f9b273cf803b85e15ba33f/google/cloud/sql/connector/connector.py#L262

the code is careful to make sure to instantiate this client in the correct async context (in the running event loop). The reason this has to be done lazily is really just because the init of CloudSQLClient initializees an aiohttp.ClientSession:

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/de2852f1e5333449f3f9b273cf803b85e15ba33f/google/cloud/sql/connector/connector.py#L262

ClientSession must be instantiated on the even loop it will be used on because of this:

https://github.com/aio-libs/aiohttp/blob/f662958b150a9d8d92fcbd0c9235e6bee1bedd67/aiohttp/client.py#L257

now, whenever a request is made with this client session, a TimeoutHandle is created, parameterized by self._loop:

https://github.com/aio-libs/aiohttp/blob/f662958b150a9d8d92fcbd0c9235e6bee1bedd67/aiohttp/client.py#L443

This propagates to a TimerContext here:

https://github.com/aio-libs/aiohttp/blob/f662958b150a9d8d92fcbd0c9235e6bee1bedd67/aiohttp/helpers.py#L651

which we enter here:

https://github.com/aio-libs/aiohttp/blob/f662958b150a9d8d92fcbd0c9235e6bee1bedd67/aiohttp/client.py#L474

now when we enter this timer context we try to find the current task:

https://github.com/aio-libs/aiohttp/blob/f662958b150a9d8d92fcbd0c9235e6bee1bedd67/aiohttp/helpers.py#L697 on the right loop, which if we have executed on a different loop, will raise an exception.

One way to fix this would be to make sure that we are running on the right event loop in _perform_refresh:

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/de2852f1e5333449f3f9b273cf803b85e15ba33f/google/cloud/sql/connector/instance.py#L148

alternatively, we could try to schedule it properly here:

[...]

PR #1113, opened 2024-06-04
fix: error on connect_async() event loop mismatch

jackwotherspoon:
WIP: Need to add proper tests to verify this.

Related to #1107

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -266,6 +266,15 @@ class Connector:
             DnsNameResolutionError: Could not resolve PSC IP address from DNS
                 host name.
         """
+        # check if event loop is running in current thread
+        if self._loop != asyncio.get_running_loop():
+            raise ConnectorLoopError(
+                "Running event loop does not match 'connector._loop'. "
+                "Connector.connect_async() must be called from the event loop "
+                "the Connector was initialized with. If you need to connect "
+                "across event loops/threads please use a new Connector object."
+            )
+
         if self._keys is None:
             self._keys = asyncio.create_task(generate_keys())
         if self._client is None:


hessjcg (approved):

kgala2 (approved):
LGTM, thanks!

palash-taneja:
i think this check is too restrictive and breaks my FastAPI application that uses SQLAlchemy and asyncpg. maybe it's better to track the loop ids instead of comparing the objects.

here's a stack trace:
https://gist.github.com/palash-taneja/5e5cf3947d95bfe886479caf8bcd8d27

hessjcg:
@palash-taneja, 

I did some research, and I have an alternate explanation: Your application likely has more than one event loop. 

The code in the connector here correctly compares event loops.  There is no possibility that the same asyncio event loop can be represented by two distinct loop objects. 

Here's the evidence from your stack trace that your app has more than one event loop: 

The first loop is runs the FastAPI router and the Cloud SQL Connector. This is the current loop when your application logs this message: "cloud_sql.py:116] [CLOUD_SQL] connect_async called from loop 5503254560, Connector was created in loop 550325456". 

Then, the code calls the SQLAlchemy code `session.execute()`. SQLAlchemy session calls `await greenlet_spawn` at `sqlalchemy/ext/asyncio/session.py:449`. This likely switches the execution from the FastAPI event loop to a different one, probably the loop that was active when the SQLAlchemy library was initialized. I would look into how the SQLAlchemy library is initialized and make sure that its event loop matches the one in the connector. I hope you figure this out.

palash-taneja:
@hessjcg where are you seeing that in the log I sent? I think there's a mistake/some clipping of the text because the log says 

"cloud_sql.py:116] [CLOUD_SQL] connect_async called from loop 5503254560, Connector was created in loop 5503254560" 

**not** "cloud_sql.py:116] [CLOUD_SQL] connect_async called from loop 5503254560, Connector was created in loop 550325456".

image

bshakur8:
In `IPython` environments, `%autoawait` manages its own event loop. To avoid 'loop mismatch' errors, fetch the active loop and pass it to the Connector. By default, the Connector [creates a new loop](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/main/google/cloud/sql/connector/connector.py#L144) if one isn't provided, which won't align with IPython's execution context.

```
from IPython.core.async_helpers import get_asyncio_loop

ipython_loop = get_asyncio_loop()

connector = Connector(loop=ipython_loop)
...
```

PR #1120, opened 2024-06-25
fix: let DNS resolve outside connector

jackwotherspoon:
Python's `SSLContext` requires trailing dot to be stripped from DNS
names in order to properly configure SSL/TLS.

Previously, we were resolving the DNS and connecting to the PSC
endpoint within the Connector. However, this was a poor workaround
as we should really just strip the trailing dot and allow the drivers to
resolve the DNS on connection.


jackwotherspoon:
Will manually test as we currently don't have integration test support for PSC.

PR #1125, opened 2024-07-08
fix: retry 50x errors with exponential backoff

jackwotherspoon:
This commit adds retry behavior to the two SQL Admin API calls. 

Any response that results in a 50x error will now be retried up to
5 times with exponential backoff and jitter between each attempt.

The formula used to calculate the duration to wait is:

200ms * 1.618^(attempt + jitter)

This calculation matches what the Cloud SQL Proxy v1 did and
will not trigger any significant change in load on the backend.


jackwotherspoon:
Going to add a unit test that covers `retry_50x` with a `CloudSQLClient`

enocom (approved):

PR #1138, opened 2024-07-29
ci: update integration test logic

jackwotherspoon:
Pull request from forks once merged will still be skipped `github.event.pull_request.head.repo.full_name` will still be that of the fork.

To combat this we run the job on all events that are not a pull_request (aka `schedule` and `push`)


enocom (approved):
A comment explaining why we're doing this would be nice for future readers.

Also, bonus points if we can split this across multiple lines to make it more readable.

PR #1156, opened 2024-09-04
test: update tests to be more scalable

jackwotherspoon:
Current integration tests use a bad practice of initializing the `Connector`
within the `getconn` creator func. This creates a `Connector` on each
database conn instead of re-using a single connector:

```python
def getconn() -> pg8000.dbapi.Connection:
    # initialize Connector object for connections to Cloud SQL
    with Connector() as connector:
```

This PR also makes a single test file be able to have tests connect to
different Cloud SQL instances (wanted for CAS tests) by removing
the pytest fixture in place of helper func.


enocom (approved):
:shipit:  Thanks for cleaning these up.

PR #1158, opened 2024-09-10
fix: set PSC ip type based on `pscEnabled`

jackwotherspoon:
With CAS-based instances being supported soon and also using `dnsName`, 
we can no longer rely on the presence of `dnsName` in the connectSettings
API as identifying that PSC is enabled on an instance. 

Instead we should check `pscEnabled` from the response as source of truth.

Check if `pscEnabled` is `True` before setting PSC ip type.


jackwotherspoon:
Need to add `pscEnabled` to int test API response.

enocom (approved):

PR #1165, opened 2024-09-17
refactor: allow cert chain for server cert

jackwotherspoon:
CAS-based CA instances can have a chain of trust.

Should allow server CA to be a list of certs to allow this cert chain.

This was part of #1154 that was closed.

### DIFF
--- a/google/cloud/sql/connector/connection_info.py
+++ b/google/cloud/sql/connector/connection_info.py
@@ -17,7 +17,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 import logging
 import ssl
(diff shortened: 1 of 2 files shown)


jackwotherspoon:
it is slightly interesting that the integration for this already was passing. I guess having an intermediate cert and chain of trust is maybe optional? 

@hessjcg do you know if chain of trust is optional here?

jackwotherspoon:
From chatting with @feng-zhe offline I now realize that this will not be a list and will still be a single string, no change is needed in the library to handle the chain of trust.

PR #1170, opened 2024-10-02
fix: remove `ConnectorLoopError` from connector.connect

### LINKED ISSUE #969: Add support for gevent
### Question

Hello,

I have a Flask app (using Flask-SqlAlchemy extension) that I am serving with Gunicorn.

If I try to use the connector (in the way it's defined for Flask-Sqlalchemy section, see attached code) using the worker class "_gevent_" I get the following error:

`google.cloud.sql.connector.exceptions.ConnectorLoopError: Connector event loop is running in current thread! Event loop must be attached to a different thread to prevent blocking code!`

However, using "_gthread_" (threads from the OS, if I am not wrong) and thus not monkey patching the code gives me no problems and the app starts properly (I guess it's because there's no event loop, so it's fine).

I'd like to know whether it's me doing something incorrect or if maybe it's because this connector is incompatible with monkey patching the code and using greenlet.

Thank you!

### Code

```Python
def getconn():
    conn = connector.connect(
        "FOO",
        "pg8000",
        user="MYUSER",
        enable_iam_auth=True,
        db="MYDB",
    )
    return conn
```

### Additional Details

_No response_

jackwotherspoon:
Python packages like [gevent](https://pypi.org/project/gevent/) have monkey patching abilities and use greenlet
processes to replace synchronous code in an "asynchronous" fashion.

The `ConnectorLoopError` in `connector.connect` was put in as a safe-guard to try
and prevent improper async driver usage where users would accidentally call
`connector.connect` for `asyncpg` instead of calling `connector.connect_async`, 
leading to an indefinite hang.

However, this error is preventing support for `gevent` which schedules tasks in
its greenlet threads.

Closes #969


enocom (approved):
Looks great! I confirmed this code works as expected when using gevent.

hessjcg (approved):

PR #1172, opened 2024-10-03
feat: improve key for connector cache

### LINKED ISSUE #1013: Improve caching of `Instance()` configs
Currently we cache `Instance()` configs by instance connection name. Which means we throw an error if users try to connect to the same instance with `enable_iam_auth=False` and then with `enable_iam_auth=True` or vice versa. Instead if we hash with the additional iam info we could support both without needing to throw an error.

Alternative: If IAM authn flag has been changed, we force a refresh to get a new cert to avoid duplicating admin API calls.

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/a8481e01ee12647743ec585c87e39580b252ef9b/google/cloud/sql/connector/connector.py#L222-L231

jackwotherspoon:
Improve dictionary key for `connector._cache`.

Previously, key for dictionary was the Cloud SQL instance's connection name
(`instance_connection_string`). 

```python
cache = connector._cache[instance_connection_string]
```
However, there is differing logic for connecting with `enable_iam_auth=True`
vs without. This resulted in having to throw an error and having a customer
create two separate connector objects if they want to connect to a single
instance with both auto IAM AuthN and regular user/password.

This PR makes the dictionary key a tuple consisting of `(instance_connection_string, enable_iam_auth)`.
This allows a single connector to cache two separate entries for the same Cloud SQL instance,
one for `enable_iam_auth=True` and one for `enable_iam_auth=False`.

```python
cache = connector._cache[(instance_connection_string, enable_iam_auth)]
```

Closes #1013

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -21,7 +21,7 @@ from functools import partial
 import logging
 from threading import Thread
 from types import TracebackType
-from typing import Any, Dict, Optional, Type, Union
+from typing import Any, Dict, Optional, Tuple, Type, Union
 
 import google.auth
 from google.auth.credentials import Credentials
@@ -131,7 +131,11 @@ class Connector:
                     asyncio.run_coroutine_threadsafe(generate_keys(), self._loop),
                     loop=self._loop,
                 )
-        self._cache: Dict[str, Union[RefreshAheadCache, LazyRefreshCache]] = {}
+        # initialize dict to store caches, key is a tuple consisting of instance
+        # connection name string and enable_iam_auth boolean flag
+        self._cache: Dict[
+            Tuple[str, bool], Union[RefreshAheadCache, LazyRefreshCache]
+        ] = {}
         self._client: Optional[CloudSQLClient] = None
 
         # initialize credentials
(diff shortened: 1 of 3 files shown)


MengmengZhangGoogle (approved):

hessjcg (approved):
Nice work, this looks good. 

A couple of ideas to for additional work:

- We need do to this for the Go connector, maybe the Node connector too, I haven't checked yet. 
- Let's consider if we should add all relevant configuration values to the key similar to how the Java connector does it. See [ConnectorConfig.java](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/ConnectorConfig.java#L28) 
  - target priniciple & delegates
  - admin url
  - quota project
  - universe domain
  - refresh strategy
  - google credentials

jackwotherspoon:
> A couple of ideas to for additional work:
> 
> * We need do to this for the Go connector, maybe the Node connector too, I haven't checked yet.
> * Let's consider if we should add all relevant configuration values to the key similar to how the Java connector does it. See [ConnectorConfig.java](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/core/src/main/java/com/google/cloud/sql/ConnectorConfig.java#L28)
>   
>   * target priniciple & delegates
>   * admin url
>   * quota project
>   * universe domain
>   * refresh strategy
>   * google credentials

@hessjcg Most (if not all) of these do not make sense given how the Python Connector works. It comes down to `Connector` level config vs `Instance` level config. Since `enable_iam_auth` is an instance level config, it makes sense to cache different entries in the cache.

However, for Connector level config attributes (credentials, HTTP client, etc), it should in fact be recommended to use a separate connector object. The `Connector()` initializes a single async HTTP client that is shared across all connections to all instances. So for things like admin url, quota project, universe domain, etc. We can not be switching them from one instance to the other.

PR #1176, opened 2024-10-17
feat: drop support for Python 3.8

jackwotherspoon:
Removing Python 3.8 support as it has reached its end of life.

[Python 3.8 EOL was October 7, 2024](https://endoflife.date/python)


sanmahapatra (approved):
LGTM

Issue #1199, opened 2024-11-22
`Service Usage Consumer` permission is required but missing from error message


uriyyo:
### Bug Description

When I try to connect to a Cloud SQL instance using the `google.cloud.sql.connector.Connector` I'm getting error:
`Forbidden: Authenticated IAM principal does not seem authorized to make API request. Verify 'Cloud SQL Admin API' is enabled within your GCP project and 'Cloud SQL Client' role has been granted to IAM principal.`

But I have already granted the `Cloud SQL Client` role to the service account. Actually, the issue is with the `Service Usage Consumer` permission. Under the hood, the `google.cloud.sql.connector.Connector` uses the `Service Usage Consumer` permission to get service metadata. So, you need to grant the `Service Usage Consumer` permission to the service account to fix this issue.

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/d622575cab34c0dc85763076f7c404e7265c3f26/google/cloud/sql/connector/client.py#L258-L262

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/d622575cab34c0dc85763076f7c404e7265c3f26/google/cloud/sql/connector/client.py#L128-L131

Here we override exception message:
https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/d622575cab34c0dc85763076f7c404e7265c3f26/google/cloud/sql/connector/instance.py#L136-L137

Because of this issue, the error message is misleading and confusing (I spent a lot of time debugging this issue 😅 ).

### Example code (or command)

```Python
import pymysql.connections
import sqlalchemy
from google.cloud.sql.connector import Connector

connector = Connector()


def getconn() -> pymysql.connections.Connection:
    conn: pymysql.connections.Connection = connector.connect(
        "project:region:instance",
        "pymysql",
        user="my-user",
        password="my-password",
        db="my-db-name",
        enable_iam_auth=True,
    )
    return conn


pool = sqlalchemy.create_engine(
    "mysql+pymysql://",
    creator=getconn,
)
```

### Steps to reproduce?

1. Try to connect to a Cloud SQL instance with SQL Admin API enabled and Cloud SQL Client role, but without Service Usage Consumer role.

### Environment

1. OS type and version: macOS 14.6.1
2. Python version: 3.12 
3. Cloud SQL Python Connector version: 1.14.0

uriyyo:
Hi @jackwotherspoon,

[...]

PR #362, opened 2022-05-19
chore: update ClientResponse error with detailed message

jackwotherspoon:
Updating error messages on both Cloud SQL API calls for `403` codes:

Previous msg:
```
aiohttp.client_exceptions.ClientResponseError: 403, message='Forbidden', url=URL('https://sqladmin.googleapis.com/sql/v1beta4/projects/project/instances/test-instance:generateEphemeralCert')
```

Updated msg:
```
aiohttp.client_exceptions.ClientResponseError: 403, message="Forbidden: Authenticated IAM principal does not seeem authorized to make API request. Verify 'Cloud SQL Admin API' is enabled within your GCP project and 'Cloud SQL Client' role has been granted to IAM principal.", url=URL('https://sqladmin.googleapis.com/sql/v1beta4/projects/project/instances/test-instance:generateEphemeralCert')
```

PR #1201, opened 2024-11-27
feat: improve aiohttp client error messages

### LINKED ISSUE #477: Improve `aiohttp` error handling by converting response body to error
The Python Connector currently uses `raise_for_status=True` when using aiohttp's [`ClientSession`](https://docs.aiohttp.org/en/stable/client_reference.html). 

The default behaviour for this argument is to close http connections when response status code > 400 and raise generic error message (ex. `400 Bad Request`), so the response body which may have a more helpful error message or instructions cannot be accessed.

It is recommended to not use `raise for status` if access to the errors response body is required. Because of this the Python Connector should do something along the following (pseudocode):

```python

def convert_response(resp):
  # attempt to read response and format custom error
  try: 
    resp_body = await resp.json()
    # ... raise error using standard response format
    build_error(resp.status, resp_body)
  # if error occurs reading response body, fallback to raise_for_status 
  except:
    resp.raise_for_status()

# --------------------------------------------------------------------
# ... inside _get_metadata (refresh_utils.py)
async with client_session.get(url, headers=headers) as resp:
  # if not 2XX status code
  if not resp.ok:
    # handle standard response and format it into error
    convert_response(resp)
  # ... continue onwards
```

jackwotherspoon:
By default, `aiohttp` on bad requests will discard response body and
output generic error message in response, `Bad Request`, `Forbidden`, etc.

The Cloud SQL Admin APIs response body contains more detailed error messages.
We need to raise these to the end user for them to be able to resolve common config 
issues on their own.

In https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/362 we caught `403` errors and implemented a custom error message
to point users at the most common config issues, Cloud SQL API not being enabled,
missing `Cloud SQL Client` role etc. However, this was not a catch-all (as noted by #1199) 
and more of a temporary fix.

This PR implements a more robust solution, copying the actual Cloud SQL Admin API
response body's error message to the end user.

> [!NOTE]
> Example common scenario: Calling the Python Connector with a typo in your instance name.

Error Message today:

```
aiohttp.client_exceptions.ClientResponseError: 404, message='Not Found'
```

Error Message after PR:

```
aiohttp.client_exceptions.ClientResponseError: 404, message='The Cloud SQL instance does not exist.'
```

Related https://github.com/aio-libs/aiohttp/issues/4600
I looked into using a custom response class but it seemed over the top
compared to the working approach here.

Closes #477

### DIFF
--- a/google/cloud/sql/connector/instance.py
+++ b/google/cloud/sql/connector/instance.py
@@ -22,8 +22,6 @@ from datetime import timedelta
 from datetime import timezone
 import logging
 
-import aiohttp
-
 from google.cloud.sql.connector.client import CloudSQLClient
 from google.cloud.sql.connector.connection_info import ConnectionInfo
 from google.cloud.sql.connector.connection_name import _parse_instance_connection_name
@@ -127,15 +125,6 @@ class RefreshAheadCache:
                 f"['{self._conn_name}']: Current certificate "
                 f"expiration = {connection_info.expiration.isoformat()}"
             )
-
-        except aiohttp.ClientResponseError as e:
-            logger.debug(
-                f"['{self._conn_name}']: Connection info "
-                f"refresh operation failed: {str(e)}"
-            )
-            if e.status == 403:
-                e.message = "Forbidden: Authenticated IAM principal does not seem authorized to make API request. Verify 'Cloud SQL Admin API' is enabled within your GCP project and 'Cloud SQL Client' role has been granted to IAM principal."
-            raise
 
         except Exception as e:
             logger.debug(
--- a/tests/unit/test_client.py
+++ b/tests/unit/test_client.py
@@ -15,6 +15,9 @@
 import datetime
 from typing import Optional
 
+from aiohttp import ClientResponseError
+from aioresponses import aioresponses
+from google.auth.credentials import Credentials
 from mocks import FakeCredentials
 import pytest
 
@@ -138,3 +141,78 @@ async def test_CloudSQLClient_user_agent(
         assert client._user_agent == f"cloud-sql-python-connector/{version}+{driver}"
     # close client
     await client.close()
+
+
+@pytest.mark.asyncio
+async def test_cloud_sql_error_messages_get_metadata(
+    fake_credentials: Credentials,
+) -> None:
+    """
+    Test that Cloud SQL Admin API error messages are raised for _get_metadata.
+    """
+    # mock Cloud SQL Admin API calls with exceptions
+    client = CloudSQLClient(
+        sqladmin_api_endpoint="https://sqladmin.googleapis.com",
+        quota_project=None,
+        credentials=fake_credentials,
+    )
+    get_url = "https://sqladmin.googleapis.com/sql/v1beta4/projects/my-project/instances/my-instance/connectSettings"
+    resp_body = {
+        "error": {
+            "code": 403,
+            "message": "Cloud SQL Admin API has not been used in project 123456789 before or it is disabled",
+        }
+    }
+    with aioresponses() as mocked:
+        mocked.get(
+            get_url,
+            status=403,
+            payload=resp_body,
+            repeat=True,
+        )
+        try:
+            await client._get_metadata("my-project", "my-region", "my-instance")
+        except ClientResponseError as e:
+            assert e.status == 403
+            assert (
+                e.message
+                == "Cloud SQL Admin API has not been used in project 123456789 before or it is disabled"
+            )
+        finally:
+            await client.close()
+
+
+@pytest.mark.asyncio
+async def test_cloud_sql_error_messages_get_ephemeral(
+    fake_credentials: Credentials,
+) -> None:
+    """
+    Test that Cloud SQL Admin API error messages are raised for _get_ephemeral.
+    """
+    # mock Cloud SQL Admin API calls with exceptions
+    client = CloudSQLClient(
+        sqladmin_api_endpoint="https://sqladmin.googleapis.com",
+        quota_project=None,
+        credentials=fake_credentials,
+    )
+    post_url = "https://sqladmin.googleapis.com/sql/v1beta4/projects/my-project/instances/my-instance:generateEphemeralCert"
+    resp_body = {
+        "error": {
+            "code": 404,
+            "message": "The Cloud SQL instance does not exist.",
+        }
+    }
+    with aioresponses() as mocked:
+        mocked.post(
+            post_url,
+            status=404,
+            payload=resp_body,
+            repeat=True,
+        )
+        try:
+            await client._get_ephemeral("my-project", "my-instance", "my-key")
+        except ClientResponseError as e:
+            assert e.status == 404
+            assert e.message == "The Cloud SQL instance does not exist."
+        finally:
+            await client.close()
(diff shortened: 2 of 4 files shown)


tests/unit/test_client.py
```diff
@@ -138,3 +141,76 @@ async def test_CloudSQLClient_user_agent(
...
+            repeat=True,
+        )
+        try:
+            await client._get_metadata("my-project", "my-region", "my-instance")
+        except ClientResponseError as e:
+            assert e.status == 403
+            assert (
+                e.message
+                == "Cloud SQL Admin API has not been used in project 123456789 before or it is disabled"
+            )
```
rhatgadkar-goog, line 179: Will this test pass if `_get_metadata()` doesn't raise an exception? We want `_get_metadata()` to raise an exception. So should we do something like?:
```
with pytest.raises(ClientResponseError) as exc_info:
    assert exc_info.status == 403
    assert exc_info.message == "Cloud SQL Admin API has not been used in project 123456789 before or it is disabled"
```

Same for `test_cloud_sql_error_messages_get_ephemeral`

jackwotherspoon: Good catch! Updated to use pytest.raises 😄

google/cloud/sql/connector/instance.py
```diff
@@ -128,15 +126,6 @@ async def _perform_refresh(self) -> ConnectionInfo:
...
-                f"['{self._conn_name}']: Connection info "
-                f"refresh operation failed: {str(e)}"
-            )
-            if e.status == 403:
-                e.message = "Forbidden: Authenticated IAM principal does not seem authorized to make API request. Verify 'Cloud SQL Admin API' is enabled within your GCP project and 'Cloud SQL Client' role has been granted to IAM principal."
-            raise
-
         except Exception as e:
             logger.debug(
                 f"['{self._conn_name}']: Connection info "
```
rhatgadkar-goog, line 131: Do we have a unit test that covers this code block? I'm guessing `test_AutoIAMAuthNotSupportedError` covers it

rhatgadkar-goog (approved):

google/cloud/sql/connector/instance.py
jackwotherspoon: Yes the coverage report shows that these lines are covered. 

Is there something specific you want tested, I don't see much value in adding specific tests around purely a debug log message but open to discussion about it.

rhatgadkar-goog: No, I didn't want anything specific tested. I just wanted to check if we had a testcase that enters that code block.

Is this the coverage report?: https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/actions/runs/12126281339/job/33808046440?pr=1201#step:5:135. And does it show that `instance.py` has 99% code coverage in tests?: BYjy7hiqre64qFv

jackwotherspoon: Yup that is the coverage report and exactly we are 99% covered for `instances.py` you can see the same report on the AlloyDB Connector as well. 

FYI: If you run nox on your local machine while working on a PR like `nox unit-3.12`, you can then see the corresponding coverage by running `coverage report --show-missing` 👍

rhatgadkar-goog: Awesome! Thank you for the info

tests/unit/test_client.py
```diff
@@ -138,3 +141,72 @@ async def test_CloudSQLClient_user_agent(
...
+            payload=resp_body,
+            repeat=True,
+        )
+        with pytest.raises(ClientResponseError) as exc_info:
+            await client._get_metadata("my-project", "my-region", "my-instance")
+            assert exc_info.status == 403
+            assert (
+                exc_info.message
+                == "Cloud SQL Admin API has not been used in project 123456789 before or it is disabled"
+            )
```
rhatgadkar-goog, line 178: These assert statements in lines 174-178 won't be executed. I was playing around with this when implementing this change for AlloyDB: https://github.com/GoogleCloudPlatform/alloydb-python-connector/pull/400.

When an exception is raised in a `pytest.raises` block, the statement that raises the exception must be the final statement in the block. More info from the pytest documentation: 
pdXPZFZZr6Uvykx


So the assert statements should go after the block, like this:
```
with pytest.raises(ClientResponseError) as exc_info:
    await client._get_metadata(...)
assert exc_info.value.status == 404
```

tests/unit/test_client.py
```diff
@@ -138,3 +141,72 @@ async def test_CloudSQLClient_user_agent(
...
+    with aioresponses() as mocked:
+        mocked.get(
+            get_url,
+            status=403,
+            payload=resp_body,
+            repeat=True,
+        )
+        with pytest.raises(ClientResponseError) as exc_info:
+            await client._get_metadata("my-project", "my-region", "my-instance")
+            assert exc_info.status == 403
```
rhatgadkar-goog, line 174: This needs to be `exc_info.value.status == 403`. Similar below: it should be `exc_info.value.message == "..."`.

`exc_info` is an `ExceptionInfo` object. To get the actual `ClientResponseError` object from it, we get it from `exc_info.value`. More info: 
8VjETmTRMP2hDfP

rhatgadkar-goog (changes requested):
I found some issues with the unit tests when I was implementing this change for AlloyDB.

al0932:
Unsubscribe On Dec 2, 2024, at 9:43 PM, rhatgadkar-goog ***@***.***> wrote:﻿
@rhatgadkar-goog commented on this pull request.



In tests/unit/test_client.py:
> +        }
+    }
+    with aioresponses() as mocked:
+        mocked.get(
+            get_url,
+            status=403,
+            payload=resp_body,
+            repeat=True,
+        )
+        with pytest.raises(ClientResponseError) as exc_info:
+            await client._get_metadata("my-project", "my-region", "my-instance")
+            assert exc_info.status == 403
+            assert (
+                exc_info.message
+                == "Cloud SQL Admin API has not been used in project 123456789 before or it is disabled"
+            )

These assert statements in lines 174-178 won't be executed. I was playing around with this when implementing this change for AlloyDB: GoogleCloudPlatform/alloydb-python-connector#400.
When an exception is raised in a pytest.raises block, the statement that raises the exception must be the final statement in the block. More info from the pytest documentation:
pdXPZFZZr6Uvykx.png (view on web)
So the assert statements should go after the block, like this:
with pytest.raises(ClientResponseError) as exc_info:
    await client._get_metadata(...)
assert exc_info.value.status == 404

tests/unit/test_client.py
jackwotherspoon: Another good catch @rhatgadkar-goog 👏 

That matches what I have in another test

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/968b6b2b5691d00edd164fdc09a3054faccbc2c6/tests/unit/test_connector.py#L88-L95

jackwotherspoon:
> I found some issues with the unit tests when I was implementing this change for AlloyDB.

@rhatgadkar-goog Moved asserts out of `pytest.raises` block 😄

rhatgadkar-goog (approved):

hessjcg (approved):

Issue #1169, opened 2024-10-01
Automatically configure and fail-over connections using DNS


jackwotherspoon:
The Python Connector may be configured to connect to a Cloud SQL instance using a DNS name.

When a connection is configured with a DNS name (instead of an instance name), the connector will query the local DNS for a TXT record. The TXT should contain an instance name matching that DNS name. The connector will use the instance name from the TXT record to connect to the Cloud SQL instance.

The connector will periodically query that DNS. If the instance name in the DNS record changes, the connector will terminate all open connections to the old instance. Then, it will reconfigure itself to open connections to the new instance.

This enables several DNS-related features by allowing configuration through DNS.

- Users can configure their database clients using the domain name instead of the instance name.
- Users can fail-over database clients when they change the DNS entry for the database instance.

PR #1204, opened 2024-11-28
feat: add support for DNS names with Connector

jackwotherspoon:
The Connector may be configured to use a DNS name to look up the instance
name instead of configuring the connector with the instance connection name directly.

Add a DNS TXT record for the Cloud SQL instance to a **private** DNS server 
or a private Google Cloud DNS Zone used by your application.  For example:

- Record type: `TXT` 
- Name: `prod-db.mycompany.example.com` – This is the domain name used by the application
- Value: `my-project:my-region:my-instance` – This is the instance connection name

Configure the Connector to use a DNS name via setting `resolver` argument to `DnsResolver`:

```python
from google.cloud.sql.connector import Connector, DnsResolver

connector = Connector(resolver=DnsResolver)
```

Call `connector.connect` with DNS domain name instead of instance connection name:

```python
conn = connector.connect(
    "prod-db.mycompany.example.com",  # using DNS domain name
    "pymysql",
     # ... insert other kwargs
)
```

Part of #1169


hessjcg (approved):
Nice work, no notes.

PR #1209, opened 2024-12-04
refactor: add `domain_name` to `ConnectionName` class

jackwotherspoon:
Add `domain_name` to `ConnectionName` dataclass.

PR also updates `_parse_connection_name` to wrap new
`_parse_connection_name_with_domain_name` to account for
domain name.

Part of #1169

### DIFF
--- a/tests/unit/test_connection_name.py
+++ b/tests/unit/test_connection_name.py
@@ -15,6 +15,9 @@
 import pytest  # noqa F401 Needed to run the tests
 
 from google.cloud.sql.connector.connection_name import _parse_connection_name
+from google.cloud.sql.connector.connection_name import (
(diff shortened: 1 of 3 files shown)


tests/unit/test_connection_name.py
```diff
@@ -14,20 +14,37 @@
 
 import pytest  # noqa F401 Needed to run the tests
 
-from google.cloud.sql.connector.connection_name import _parse_instance_connection_name
+# fmt: off
```
jackwotherspoon, line 17: import style of long import names differs between `isort` and `black` format, turning black formatter off for these imports to stop lint job from complaining.

hessjcg (approved):

PR #1212, opened 2024-12-07
refactor: add `ConnectionName` to `ConnectionInfo` class

jackwotherspoon:
Adding `ConnectionName` to `ConnectionInfo` class to more closely match [Go implementation](https://github.com/GoogleCloudPlatform/cloud-sql-go-connector/blob/5061c61092c3fa92175ca8502a365e6e82993b6e/internal/cloudsql/instance.go#L175).

Benefits:

- Cleaner interface, passing connection name around instead of individual args (project, region, instance)
- Consistent debug logs (give access to ConnectionName throughout)
- Allows us to tell if ConnectionInfo is using DNS  (if `ConnectionName.domain_name` is set)

This will unblock DNS polling as we must know at the `ConnectionInfo` level if it is
configured via DNS domain name or not.

Related to #1169

### DIFF
--- a/google/cloud/sql/connector/client.py
+++ b/google/cloud/sql/connector/client.py
@@ -26,6 +26,7 @@ from google.auth.credentials import TokenState
 from google.auth.transport import requests
 
 from google.cloud.sql.connector.connection_info import ConnectionInfo
+from google.cloud.sql.connector.connection_name import ConnectionName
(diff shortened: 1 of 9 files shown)


jackwotherspoon:
Lint will pass once we merge https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/1211 first

hessjcg (approved):

PR #1231, opened 2025-01-31
feat: add support for `GOOGLE_CLOUD_QUOTA_PROJECT` env var

### LINKED ISSUE #1229: Add support for `GOOGLE_CLOUD_QUOTA_PROJECT` env var
We should add support for reading `GOOGLE_CLOUD_QUOTA_PROJECT` from the environment as an alternative way of setting the quota project to be used for Cloud SQL Admin API calls.

Currently we only support quota project via the `quota_project` argument to `Connector()`:

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/144311b715aedea467a4cc1a783ea60e6b6ad27a/google/cloud/sql/connector/connector.py#L64

Similar to [`GOOGLE_CLOUD_UNIVERSE_DOMAIN`](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/1221) support that we added.

jackwotherspoon:
Add support for `quota_project` to be set via the `GOOGLE_CLOUD_QUOTA_PROJECT` env var.

Order of operations for setting quota project will be as follows:

1. Use `quota_project` argument passed to `Connector()`
2. Use `GOOGLE_CLOUD_QUOTA_PROJECT` env var
3. Set default to `None` (project set in environment)

Fixes #1229

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -165,13 +165,17 @@ class Connector:
         # set default params for connections
         self._timeout = timeout
         self._enable_iam_auth = enable_iam_auth
-        self._quota_project = quota_project
         self._user_agent = user_agent
         self._resolver = resolver()
         # if ip_type is str, convert to IPTypes enum
         if isinstance(ip_type, str):
             ip_type = IPTypes._from_str(ip_type)
         self._ip_type = ip_type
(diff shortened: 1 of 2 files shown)


kgala2 (approved):
LGTM

PR #1236, opened 2025-02-12
build: update release.sh to not use setup.py

jackwotherspoon:
In #1230 we moved package off of `setup.py` and over to `pyproject.toml`

Forgot a spot to update in `release.sh` which references `setup.py` which no longer exists.

Related to #1230


kgala2 (approved):
LGTM

PR #1238, opened 2025-02-19
chore: audit test csql instances

kgala2:
Consolidate Cloud SQL instance usage for end-to-end connectivity testing in the following cases:
1. MySQL IAM connection 
2. Postgres IAM connection


CONTRIBUTING.md
```diff
@@ -73,7 +73,7 @@ export POSTGRES_PASS="db_pass"
 export POSTGRES_DB="db_name"
 # Postgres instance with IAM authentication enabled
 # (can be same as POSTGRES_CONNECTION_NAME)
-export POSTGRES_IAM_CONNECTION_NAME="<PROJECT>:<REGION>:<INSTANCE>"
+export POSTGRES_CONNECTION_NAME="<PROJECT>:<REGION>:<INSTANCE>"
```
jackwotherspoon, line 76: we should just remove this line as we already have an export line for this variable above on line 70...

We should copy the comment `# Postgres instance with IAM authentication enabled` to above line 70.

jackwotherspoon: Also should do the same for `MYSQL_IAM_CONNECTION_NAME`

jackwotherspoon (changes requested):

CONTRIBUTING.md
```diff
@@ -61,19 +61,13 @@ export MYSQL_CONNECTION_NAME="<PROJECT>:<REGION>:<INSTANCE>"
 export MYSQL_USER="db_user"
 export MYSQL_PASS="db_pass"
 export MYSQL_DB="db_name"
-# MySQL instance with IAM authentication enabled
```
jackwotherspoon, line 64: can we move the `# MySQL instance with IAM authentication enabled` comment above
```sh
export MYSQL_CONNECTION_NAME="<PROJECT>:<REGION>:<INSTANCE>"
```

Same with Postgres below...

jackwotherspoon (changes requested):

PR #1241, opened 2025-03-03
feat: reset connection when the DNS record changes

### LINKED ISSUE #1169: Automatically configure and fail-over connections using DNS

jackwotherspoon:
If the connector is configured with a domain name, when that domain name record changes to
resolve to a new instance, the connector should detect that change, close all connections to the old
instance, and create connections to the new instance.

```python
import sqlalchemy
from google.cloud.sql.connector import Connector, DnsResolver

# using default failover period of 30 seconds
connector = Connector(resolver=DnsResolver)

# using custom failover period
# connector = Connector(resolver=DnsResolver, failover_period=60)

# establish connection using domain name
engine = sqlalchemy.create_engine(
    "postgresql+pg8000://",
    creator=lambda: connector.connect(
        "db.example.com",
        "pg8000",
        user="my-user",
        password="my-password",
        db="my-db",
        ip_type="public",  # can also be "private" or "psc"
    ),
)
with engine.connect() as conn:
      curr_time = conn.execute(sqlalchemy.text("SELECT NOW()")).fetchone()
      conn.commit()
```

If `db.example.com` fails over from `my-project:us-central1:pg` to `my-project:us-central1:pg-replica` then the connections will be closed and re-opened to new instance.

```bash
2025-03-20 14:07:05,504 [DEBUG]: ['db.example.com -> my-project:us-central1:pg']: Connecting to 34.172.89.123:3307
2025-03-20 14:07:16,137 [DEBUG]: ['db.example.com -> my-project:us-central1:pg-replica']: Connecting to 34.60.127.180:3307
2025-03-20 14:07:19,387 [DEBUG]: ['db.example.com -> my-project:us-central1:pg']: Cloud SQL instance changed from my-project:us-central1:pg to my-project:us-central1:pg-replica, closing all connections!
2025-03-20 14:07:19,387 [DEBUG]: ['db.example.com -> my-project:us-central1:pg']: Cancelled domain name polling task.
2025-03-20 14:07:19,387 [DEBUG]: ['db.example.com -> my-project:us-central1:pg']: Canceling connection info refresh operation tasks
2025-03-20 14:07:19,387 [DEBUG]: ['db.example.com -> my-project:us-central1:pg']: Scheduled refresh operation cancelled
2025-03-20 14:07:21,472 [DEBUG]: ['db.example.com -> my-project:us-central1:pg-replica']: Connecting to 34.60.127.180:3307
```

Closes #1169


google/cloud/sql/connector/connection_name.py
```diff
@@ -38,6 +38,10 @@ def __str__(self) -> str:
             return f"{self.domain_name} -> {self.project}:{self.region}:{self.instance_name}"
         return f"{self.project}:{self.region}:{self.instance_name}"
 
+    def get_connection_string(self) -> str:
+        """Get the instance connection string for the Cloud SQL instance."""
+        return f"{self.project}:{self.region}:{self.instance_name}"
+
```
jackwotherspoon, line 44: This allows for a reliable way to get the instance connection name for a Cloud SQL instance whether the connector is connecting via domain name or instance connection name.

google/cloud/sql/connector/connector.py
```diff
@@ -285,15 +293,16 @@ async def connect_async(
                 driver=driver,
             )
         enable_iam_auth = kwargs.pop("enable_iam_auth", self._enable_iam_auth)
-        if (instance_connection_string, enable_iam_auth) in self._cache:
-            cache = self._cache[(instance_connection_string, enable_iam_auth)]
+
+        conn_name = await self._resolver.resolve(instance_connection_string)
+        if (str(conn_name), enable_iam_auth) in self._cache:
+            monitored_cache = self._cache[(str(conn_name), enable_iam_auth)]
```
jackwotherspoon, line 299: Update cache key to be `str(conn_name)` which will result in cache key either being :

`domain name -> instance connection name` or just `instance connection name`

jackwotherspoon:
Code coverage drop from 94% to 93% is negligible

google/cloud/sql/connector/monitored_cache.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Close any still open sockets
+        for socket in self.sockets:
+            # Check fileno as method to check if socket is closed. Will return
+            # -1 on failure, which will be used to signal socket closed.
+            if socket.fileno() != -1:
+                socket.close()
+
+        # Close underyling ConnectionInfoCache
+        await self.cache.close()
```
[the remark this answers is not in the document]
jackwotherspoon: Good catch, updated!

google/cloud/sql/connector/connector.py
```diff
@@ -285,15 +293,16 @@ async def connect_async(
                 driver=driver,
             )
         enable_iam_auth = kwargs.pop("enable_iam_auth", self._enable_iam_auth)
-        if (instance_connection_string, enable_iam_auth) in self._cache:
-            cache = self._cache[(instance_connection_string, enable_iam_auth)]
+
+        conn_name = await self._resolver.resolve(instance_connection_string)
+        if (str(conn_name), enable_iam_auth) in self._cache:
```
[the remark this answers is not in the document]
jackwotherspoon: Updated!

google/cloud/sql/connector/instance.py
```diff
@@ -74,6 +75,15 @@ def __init__(
...
+
+    @property
+    def conn_name(self) -> ConnectionName:
+        return self._conn_name
+
+    @property
+    def closed(self) -> bool:
+        return self._closed
 
     async def force_refresh(self) -> None:
```
hessjcg, line 88: This should probably balk if the instance is closed: 
```python
async def force_refresh(self) -> None:
    if self._closed: 
        return
```
Same for `_perform_refresh(self)` and `connect_info(self)`

jackwotherspoon: I put these on the MonitoredCache class to not have to do on both background and lazy caches.

hessjcg (approved):

PR #1255, opened 2025-03-28
test: remove test_connector_with_credentials system test

kgala2:
Removing test_connector_with_credentials from tests/system/test_connector_object.py as unit tests already cover the credential tests

Reference bug: b/396351603


jackwotherspoon (approved):
LGTM ✅

PR #1256, opened 2025-03-28
test: update ip type for connectors

kgala2:
Updates ip_type parameter to use environment variable - "IP_TYPE" to enable private IP tests. The default still remains "public"

Reference: b/396351603

### DIFF
--- a/tests/system/test_asyncpg_connection.py
+++ b/tests/system/test_asyncpg_connection.py
@@ -32,7 +32,7 @@ async def create_sqlalchemy_engine(
     user: str,
     password: str,
     db: str,
-    ip_type: str,
+    ip_type: str = "public",
     refresh_strategy: str = "background",
     resolver: Union[type[DefaultResolver], type[DnsResolver]] = DefaultResolver,
 ) -> tuple[sqlalchemy.ext.asyncio.engine.AsyncEngine, Connector]:
@@ -65,7 +65,7 @@ async def create_sqlalchemy_engine(
         db (str):
             The name of the database, e.g., mydb
         ip_type (str):
-            The IP type of the Cloud SQL instance.
+            The IP type of the Cloud SQL instance. Can be one of "public", "private", or "psc".
         refresh_strategy (Optional[str]):
             Refresh strategy for the Cloud SQL Connector. Can be one of "lazy"
             or "background". For serverless environments use "lazy" to avoid
@@ -90,7 +90,7 @@ async def create_sqlalchemy_engine(
             user=user,
             password=password,
             db=db,
-            ip_type=ip_type,
+            ip_type=ip_type,  # can be "public", "private" or "psc"
         ),
         execution_options={"isolation_level": "AUTOCOMMIT"},
     )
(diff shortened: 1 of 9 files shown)


tests/system/test_asyncpg_connection.py
```diff
@@ -87,7 +87,9 @@ async def create_sqlalchemy_engine(
             user=user,
             password=password,
             db=db,
-            ip_type="public",  # can also be "private" or "psc"
+            ip_type=os.environ.get(
+                "IP_TYPE", "public"
+            ),  # can also be "private" or "psc"
```
jackwotherspoon, line 92: let's add `ip_type` to the `create_sqlalchemy_engine` helper function:

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/eba2228cf419827ca62162ab9be4d5c89f2dd0c2/tests/system/test_pg8000_connection.py#L30-L37

This way we can read `IP_TYPE` env var with the others in each test case and then pass it to helper function...

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/eba2228cf419827ca62162ab9be4d5c89f2dd0c2/tests/system/test_pg8000_connection.py#L99-L106

jackwotherspoon (changes requested):

tests/system/test_asyncpg_connection.py
kgala2: Done.

tests/system/test_asyncpg_connection.py
```diff
@@ -32,6 +32,7 @@ async def create_sqlalchemy_engine(
     user: str,
     password: str,
     db: str,
+    ip_type: str,
```
jackwotherspoon, line 35: nit: let's default this to `"public"` so docs samples work
```suggestion
    ip_type: str = "public",
```

tests/system/test_asyncpg_connection.py
```diff
@@ -63,6 +64,8 @@ async def create_sqlalchemy_engine(
             The database user's password, e.g., secret-password
         db (str):
             The name of the database, e.g., mydb
+        ip_type (str):
+            The IP type of the Cloud SQL instance.
```
jackwotherspoon, line 68: nit: always best to provide available options in docstring
```suggestion
        ip_type (str):
            The IP type of the Cloud SQL instance to connect to. Can be one of "public",
            "private", or "psc".
```

tests/system/test_asyncpg_connection.py
```diff
@@ -87,7 +90,7 @@ async def create_sqlalchemy_engine(
             user=user,
             password=password,
             db=db,
-            ip_type="public",  # can also be "private" or "psc"
+            ip_type=ip_type,
```
jackwotherspoon, line 93: nit: instead of this comment being during test case, it should be here so that it shows up in the docs
```suggestion
            ip_type=ip_type,  # can be "public", "private" or "psc"
```

jackwotherspoon (changes requested):
couple small items to improve docs

tests/system/test_asyncpg_connection.py
```diff
@@ -63,6 +64,8 @@ async def create_sqlalchemy_engine(
             The database user's password, e.g., secret-password
         db (str):
             The name of the database, e.g., mydb
+        ip_type (str):
+            The IP type of the Cloud SQL instance. Can be one of "public", "private", or "psc".
```
jackwotherspoon, line 68: nit: let's update this to not go over 80 chars per line

```suggestion
            The IP type of the Cloud SQL instance to connect to. Can be one of "public",
            "private", or "psc".
```

kgala2: Done.

PR #1257, opened 2025-03-28
ci: add cloudbuild config for PSA tests

kgala2:
This PR enables private IP tests for Cloud SQL instances with the help of Cloud Build private pool

Introduced a new environment variable - IP_TYPE which is set as "private" for PSA tests in the cloudbuild.yaml file 

reference: b/396351603


jackwotherspoon (approved):
LGTM 👏 ✅

PR #1269, opened 2025-04-07
chore: add check for closed connector

kgala2:
Draft-PR to check if the connector's been closed before trying to connect using the same engine.

1. Throws error if the connector has been closed 
2. Continues to connect otherwise

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -154,6 +154,7 @@ class Connector:
         # connection name string and enable_iam_auth boolean flag
         self._cache: dict[tuple[str, bool], MonitoredCache] = {}
         self._client: Optional[CloudSQLClient] = None
+        self._closed: bool = False
 
         # initialize credentials
         scopes = ["https://www.googleapis.com/auth/sqlservice.admin"]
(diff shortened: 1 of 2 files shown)


google/cloud/sql/connector/connector.py
```diff
@@ -463,13 +468,15 @@ def close(self) -> None:
                 self._loop.call_soon_threadsafe(self._loop.stop)
             # wait for thread to finish closing (i.e. loop to stop)
             self._thread.join()
+        self._closed = True
```
jackwotherspoon, line 471: this is not necessary as synchronous close calls `close_async` so should get closed by `close_async`

google/cloud/sql/connector/connector.py
```diff
@@ -473,13 +487,15 @@ def close(self) -> None:
...
 
     async def close_async(self) -> None:
         """Helper function to cancel the cache's tasks
         and close aiohttp.ClientSession."""
         await asyncio.gather(*[cache.close() for cache in self._cache.values()])
         if self._client:
             await self._client.close()
+        self._closed = True
```
hessjcg: Do this first, then client.close(), then gather cache.close()

Issue #214, opened 2021-12-10
Add support for psycopg2 driver.


jackwotherspoon:
Support PostgreSQL [psycopg2](https://github.com/psycopg/psycopg2/) driver with connector.

gunar:
Hey @jackwotherspoon, I've hit this exact problem this week. Do you have an estimate? Any rough ideas would be greatly appreciated. Thanks in advance.

jackwotherspoon:
@gunar This has been prioritized and will be worked on in early 2022. 

In the meantime, you can connect to Cloud SQL using Python and psycopg2 with the Cloud SQL Auth Proxy through Unix sockets or TCP. For an explanation and example see here: https://cloud.google.com/sql/docs/postgres/connect-admin-proxy#expandable-2

For the Unix sockets sample just swap out part of the pg8000 code for this code snippet:
```
pool = sqlalchemy.create_engine(
    sqlalchemy.engine.url.URL.create(
        drivername="postgresql+psycopg2",
        username=db_user,  # e.g. "my-database-user"
        password=db_pass,  # e.g. "my-database-password"
        database=db_name,  # e.g. "my-database-name"
        query={
            "host: "{}/{}".format(
                db_socket_dir,  # e.g. "/cloudsql"
                instance_connection_name)  # i.e "<PROJECT-NAME>:<INSTANCE-REGION>:<INSTANCE-NAME>"
        }
    ),
    **db_config
)
```

Hope this helps for the time being! Thanks for the feedback!

gunar:
@jackwotherspoon Thanks man. I'm not inclined to go with the Cloud SQL Auth Proxy as that's one more thing for the user to remember to execute.

I guess I could make the python script itself spawn the process for the proxy—has anyone attempted that?

Also, do I need `sqlalchemy` in order to connect to the proxy or can I connect directly with `psycopg2`?

kemics:
Hi @jackwotherspoon , thank you for all your work on cloud sql connectors

Do you have any estimates for psycopg2 support? It would be really nice to have

By the way the issue has p2 priority but it has the biggest number of thumb-ups😉

jackwotherspoon:
Hi @kemics, thanks for the kind words and showing your enthusiasm towards the Cloud SQL Python connector.

I currently can't provide a specific timeline for psycopg2 support because there are a lot of factors at play here but I can provide some details for you. 

[...]

Issue #219, opened 2021-12-10
Consider support for psycopg3 driver.


jackwotherspoon:
Monitor interest and adoption of [psycopg3](https://github.com/psycopg/psycopg) driver for potential support with Cloud SQL connector.

khaerulumam42:
very interesting to watch this development, thank you team!

jackwotherspoon:
Re-commenting what I commented on #214 as it also applies to psycopg3

FYI to everyone: it is possible to connect with automatic IAM authentication without the need for the Python Connector:

You'll need to ensure a few things:

- The token has only `sql.login` scope (i.e. `https://www.googleapis.com/auth/sqlservice.login`)
- The token isn't transmitted over an unencrypted channel. (Private IP recommended)

We're working on making this path easier for folks, but for now I'll share the mechanics for visibility.

Assuming you're using [SQLAlchemy](https://www.sqlalchemy.org/), you can do this:
```python
import sqlalchemy
from sqlalchemy import event

import google.auth
from google.auth.transport.requests import Request

# initialize ADC creds
creds, _ = google.auth.default(
    scopes=["https://www.googleapis.com/auth/sqlservice.login"]
)

# Cloud SQL Instance IP address
instance_ip = <INSTANCE_IP> 

# create SQLAlchemy connection pool
# use Cloud SQL Instance IP + native port (5432)
# for best security use client certificates + server cert for SSL
engine = sqlalchemy.create_engine(
    f"postgresql+psycopg://service-account@project-id.iam:empty-pass@{ip_address}:5432/dbname",
    connect_args={'sslmode': 'require'},
)

# set do_connect event listener
# This gets called before a connection is created and allows you to
# refresh the OAuth2 token here as needed
@event.listens_for(engine, 'do_connect')
def auto_iam_authn(dialect, conn_rec, cargs, cparams):  
    # refresh credentials if expired
    if not creds.valid:
        request = Request()
        creds.refresh(request)
    
    # use OAuth2 token as password
    cparams["password"] = str(creds.token)

# interact with Cloud SQL database using connection pool
with engine.connect() as conn:
    # query database
    time = conn.execute(sqlalchemy.text("SELECT NOW()")).fetchone()
    print(time[0])
```

nick-catalist:
FYI for anyone looking to make use of psycopg and who wants to continue relying on Google's Cloud SQL Auth Proxy and not deal with VPCs, there is a way!

[...]

PR #1311, opened 2025-07-16
feat(main): Add support for psycopg (Issue #214)

UzielSilva:
See https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/issues/214
Changelog:
- Add proxy for connections that can only be made through an unix socket, to support the TLS connection
- Add support for psycopg, using the proxy server
- Add unit and integration tests
- Update docs

### DIFF
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -35,10 +35,10 @@ from google.cloud.sql.connector.enums import IPTypes
 from google.cloud.sql.connector.enums import RefreshStrategy
 from google.cloud.sql.connector.instance import RefreshAheadCache
 from google.cloud.sql.connector.lazy import LazyRefreshCache
+import google.cloud.sql.connector.local_unix_socket as local_unix_socket
 from google.cloud.sql.connector.monitored_cache import MonitoredCache
 import google.cloud.sql.connector.pg8000 as pg8000
-import google.cloud.sql.connector.proxy as proxy
-import google.cloud.sql.connector.psycopg as psycopg
+from google.cloud.sql.connector.proxy import Proxy
 import google.cloud.sql.connector.pymysql as pymysql
 import google.cloud.sql.connector.pytds as pytds
 from google.cloud.sql.connector.resolver import DefaultResolver
@@ -49,7 +49,6 @@ from google.cloud.sql.connector.utils import generate_keys
 logger = logging.getLogger(name=__name__)
 
 ASYNC_DRIVERS = ["asyncpg"]
-LOCAL_PROXY_DRIVERS = ["psycopg"]
 SERVER_PROXY_PORT = 3307
 _DEFAULT_SCHEME = "https://"
 _DEFAULT_UNIVERSE_DOMAIN = "googleapis.com"
@@ -156,7 +155,7 @@ class Connector:
         # connection name string and enable_iam_auth boolean flag
         self._cache: dict[tuple[str, bool], MonitoredCache] = {}
         self._client: Optional[CloudSQLClient] = None
-        self._proxy: Optional[asyncio.Task] = None
+        self._proxies: Optional[Proxy] = None
 
         # initialize credentials
         scopes = ["https://www.googleapis.com/auth/sqlservice.admin"]
@@ -217,6 +216,29 @@ class Connector:
     def universe_domain(self) -> str:
         return self._universe_domain or _DEFAULT_UNIVERSE_DOMAIN
 
+    def start_unix_socket_proxy_async(
+        self,
+        instance_connection_name: str,
+        local_socket_path: str,
+        **kwargs: Any
+    ) -> None:
+        """Creates a new Proxy instance and stores it to properly disposal
+
+        Args:
+            instance_connection_string (str): The instance connection name of the
+                Cloud SQL instance to connect to. Takes the form of
+                "project-id:region:instance-name"
+
+                Example: "my-project:us-central1:my-instance"
+
+            local_socket_path (str): A string representing the location of the local socket.
+
+            **kwargs: Any driver-specific arguments to pass to the underlying
+                driver .connect call.
+        """
+        # TODO: validates the local socket path is not the same as other invocation
+        self._proxies.append(new Proxy(self, instance_connection_name, local_socket_path, self.loop, **kwargs))
+
     def connect(
         self, instance_connection_string: str, driver: str, **kwargs: Any
     ) -> Any:
@@ -234,7 +256,7 @@ class Connector:
                 Example: "my-project:us-central1:my-instance"
 
             driver (str): A string representing the database driver to connect
-                with. Supported drivers are pymysql, pg8000, psycopg, and pytds.
+                with. Supported drivers are pymysql, pg8000, local_unix_socket, and pytds.
 
             **kwargs: Any driver-specific arguments to pass to the underlying
                 driver .connect call.
@@ -270,7 +292,7 @@ class Connector:
                 Example: "my-project:us-central1:my-instance"
 
             driver (str): A string representing the database driver to connect
-                with. Supported drivers are pymysql, asyncpg, pg8000, psycopg, and
+                with. Supported drivers are pymysql, asyncpg, pg8000, local_unix_socket, and
                 pytds.
 
             **kwargs: Any driver-specific arguments to pass to the underlying
@@ -283,7 +305,7 @@ class Connector:
             ValueError: Connection attempt with built-in database authentication
                 and then subsequent attempt with IAM database authentication.
             KeyError: Unsupported database driver Must be one of pymysql, asyncpg,
-                pg8000, psycopg, and pytds.
+                pg8000, local_unix_socket, and pytds.
         """
         if self._keys is None:
             self._keys = asyncio.create_task(generate_keys())
@@ -337,7 +359,7 @@ class Connector:
         connect_func = {
             "pymysql": pymysql.connect,
             "pg8000": pg8000.connect,
-            "psycopg": psycopg.connect,
+            "local_unix_socket": local_unix_socket.connect,
             "asyncpg": asyncpg.connect,
             "pytds": pytds.connect,
         }
@@ -396,17 +418,6 @@ class Connector:
                 socket.create_connection((ip_address, SERVER_PROXY_PORT)),
                 server_hostname=ip_address,
             )
-
-            host = ip_address
-            # start local proxy if driver needs it
-            if driver in LOCAL_PROXY_DRIVERS:
-                local_socket_path = kwargs.pop("local_socket_path", "/tmp/connector-socket")
-                host = local_socket_path
-                self._proxy = proxy.start_local_proxy(
-                    sock,
-                    socket_path=f"{local_socket_path}/.s.PGSQL.{SERVER_PROXY_PORT}",
-                    loop=self._loop
-                )
 
             # If this connection was opened using a domain name, then store it
             # for later in case we need to forcibly close it on failover.
@@ -488,11 +499,7 @@ class Connector:
         if self._client:
             await self._client.close()
         if self._proxy:
-            proxy_task = asyncio.gather(self._proxy)
-            try:
-                await asyncio.wait_for(proxy_task, timeout=0.1)
-            except (asyncio.CancelledError, asyncio.TimeoutError, TimeoutError):
-                pass # This task runs forever so it is expected to throw this exception
+            await asyncio.wait_for([ proxy.close_async() for proxy in self._proxies])
 
 
 
--- a/google/cloud/sql/connector/connector.py
+++ b/google/cloud/sql/connector/connector.py
@@ -37,7 +37,7 @@ from google.cloud.sql.connector.instance import RefreshAheadCache
 from google.cloud.sql.connector.lazy import LazyRefreshCache
 from google.cloud.sql.connector.monitored_cache import MonitoredCache
 import google.cloud.sql.connector.pg8000 as pg8000
-from google.cloud.sql.connector.proxy import start_local_proxy
+import google.cloud.sql.connector.proxy as proxy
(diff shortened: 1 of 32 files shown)


hessjcg:
Thank you for the contribution! We are excited to add this to the code. I will give this careful review and get back in touch soon with some additional comments.

hessjcg (commented):
Nice work so far. This is going to be a very popular enhancement. 

You may want to look at a similar local unix proxy implementation in the [node connector](https://github.com/GoogleCloudPlatform/cloud-sql-nodejs-connector/blob/main/src/connector.ts#L316). 

There are a few things I'd like you to add to this PR. This will get it closer to production-worthy code:

1. The socket path should be a configuration parameter `local_socket_path` on `connector.connect()` instead of being hard-coded. This will allow an application to open many connectors to different database instances.
2. When the listener opens the unix socket, it should set the file permissions to be readable only by the user.
3. Handle exceptions from creating and binding the unix socket, log an appropriate error, and raise a LocalProxyStartupError. Define this new exception in google/cloud/sql/connector/exceptions.py
4. Consider using asyncio instead of a thread to copy data to and from the proxy. 
5. When the connector is closed, the `local_communication` loop must be stopped and the socket file removed.

Again, I'm grateful that you put in the time and wrote this feature. Thank you.

enocom:
cc @rhatgadkar-goog who owns the AlloyDB Python Connector and will be interested to see this.

Big +1 on what @hessjcg. This will be a huge win for people wanting to use psycopg and Cloud SQL (or AlloyDB).

UzielSilva:
All your comments were addressed on my last commit, I even improve the integration test to use SQLAlchemy like other tests.
I only have this issue when I ran the integration test:
```
Task was destroyed but it is pending!
task: <Task pending name='Task-12' coro=<local_communication() done, defined at /home/uziel/Enterprises/TSA/cloud-sql-python-connector/google/cloud/sql/connector/proxy.py:72> wait_for=<Future pending cb=[BaseSelectorEventLoop._sock_read_done(20, handle=<Handle BaseS...0>, 10485760)>)(), Task.task_wakeup()]>>
```
Not sure how to handle or if it is expected, this is caused because the `local_communication` is still expecting your input when you close the connection. Not sure if there is a more friendly way to stop it, but oh well.
Let me know if this is good enough for production.

UzielSilva:
FYI: The last issue I mentioned is only a warning, so, actually, it is not affecting the feature.

rhatgadkar-goog:
> All your comments were addressed on my last commit, I even improve the integration test to use SQLAlchemy like other tests. I only have this issue when I ran the integration test:
> 
> ```
> Task was destroyed but it is pending!
> task: <Task pending name='Task-12' coro=<local_communication() done, defined at /home/uziel/Enterprises/TSA/cloud-sql-python-connector/google/cloud/sql/connector/proxy.py:72> wait_for=<Future pending cb=[BaseSelectorEventLoop._sock_read_done(20, handle=<Handle BaseS...0>, 10485760)>)(), Task.task_wakeup()]>>
> ```
> 
> Not sure how to handle or if it is expected, this is caused because the `local_communication` is still expecting your input when you close the connection. Not sure if there is a more friendly way to stop it, but oh well. Let me know if this is good enough for production.

Thank you for this PR of adding support for psycopg! We really appreciate this change.

I took a look at this StackOverflow post to help understand what might be happening here: https://stackoverflow.com/questions/71542947/how-can-i-fix-task-was-destroyed-but-it-is-pending.

I think you're getting the error, because when the connector closes, you're not cancelling the local proxy server task and waiting for it to shutdown. So maybe you can try making the following changes:
- Modify `start_local_proxy()` function to return the created asyncio task and save this as a member variable in the `Connector` object.
- And then in the `Connector` class's `close_async()` function, you can cancel the proxy server's asyncio task and gracefully wait for it to cancel. [RefreshAheadCache's close() function](https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/blob/28c1c40c7b5e29f858f30d849c8b2c733d623f30/google/cloud/sql/connector/instance.py#L212) might be a good example for how to do this.

UzielSilva:
@rhatgadkar-goog It worked!, now the warning is gone and the server is being stopped gracefully. Thanks!

So, I think the PR is ready for a review, let me know if you want me to do some other modifications. :)

google/cloud/sql/connector/connector.py
```diff
@@ -380,7 +385,7 @@ async def connect_async(
             # async drivers are unblocking and can be awaited directly
             if driver in ASYNC_DRIVERS:
                 return await connector(
-                    ip_address,
```
kgala2, line 383: The `host` seems to be undefined here

google/cloud/sql/connector/connector.py
```diff
@@ -390,14 +395,26 @@ async def connect_async(
                 socket.create_connection((ip_address, SERVER_PROXY_PORT)),
                 server_hostname=ip_address,
             )
+
+            host = ip_address
+            # start local proxy if driver needs it
+            if driver in LOCAL_PROXY_DRIVERS:
+                local_socket_path = kwargs.pop("local_socket_path", "/tmp/connector-socket")
+                host = local_socket_path
+                self._proxy = start_local_proxy(
```
kgala2, line 404: can we initial the self._proxy in the constructor as it might not be defined for other drivers and will cause error in `close_aync` func

google/cloud/sql/connector/psycopg.py
```diff
@@ -0,0 +1,63 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import ssl
+from typing import Any, TYPE_CHECKING
+import threading
```
kgala2, line 19: This is an unused import, to run and correct lint errors consider - 

pip install nox 
nox -s lint

kgala2 (changes requested):

google/cloud/sql/connector/psycopg.py
UzielSilva: All linting issues addressed

google/cloud/sql/connector/connector.py
UzielSilva: `self._proxy` variable is now defined as an `Optional[asyncio.Task]` variable in the constructor and initialized as `None`

google/cloud/sql/connector/connector.py
UzielSilva: whoops, moving back to ip_address

UzielSilva:
Also fixed a unit test, test coverage is restored to 100%

UzielSilva:
Trying to add tests for the proxy server, to increase code coverage.

UzielSilva:
Tests were added, code coverage is now at 94%

UzielSilva:
Huh, do I need to also keep support for python 3.9? let me check then the fixes I need to do for it. :)

UzielSilva:
Added support for Python 3.9, only needed to add some exceptions to handle when I close the server

UzielSilva:
Still not able to merge it, it says changes requested

enocom:
Only maintainers can merge.

@kgala2 will have to merge. 

Great work on this @UzielSilva :clap:.

kgala2:
Hi @UzielSilva,

The overall PR looks good, I am checking for corner cases that might not be covered, I'll comment on this PR if I find anything and have psyopg support available as soon as possible. 
cc: @hessjcg 

Thanks

sahibpandori:
Thanks for working on this @UzielSilva, very excited for this change! Hope this can be merged soon @kgala2 @hessjcg

UzielSilva:
Wondering if I should specify the binary flavor for the psycopg optional dependency in the pyproject.toml. I'm using that one for the tests that are working well... not sure how it is going to behave with the purely python based installation.

Ref: https://www.psycopg.org/psycopg3/docs/basic/install.html

CC: @kgala2

pyproject.toml
```diff
@@ -59,6 +59,7 @@ Changelog = "https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/b
 [project.optional-dependencies]
 pymysql = ["PyMySQL>=1.1.0"]
 pg8000 = ["pg8000>=1.31.1"]
+psycopg = ["psycopg>=3.2.9"]
```
UzielSilva, line 62: So, maybe we need to update this with `psycopg[binary]`

jeffclay: What are the pro's and con's for the different approaches?

hessjcg (changes requested):
Excellent work so far @UzielSilva. This will fix both #219 and #214 

In reviewing the implementation, this is a good prototype, but it needs to handle sockets in a different way in order to be production-worthy. As implemented today, it can only handle a single connection. [Socket Programming in Python](https://realpython.com/python-sockets/#handling-multiple-connections), particularly the section on "handling multiple connections" explains how the code should handle the unix server socket.

Also, I think we should rename this feature to "Local Unix Socket" instead of "Psychopg support". We are implementing this in a way that is not database driver specific, and will work on any system that supports unix sockets. 

First, lets make a new class in `proxy.py` called `Proxy`. This will encapsulate the details for the proxy server. 

Create a new method on the Connector: `Connector.start_unix_socket_proxy_async`, which will create a new `Proxy` object. The Connector should allow `Connector.start_unix_socket_proxy_async()` to be called multiple times, opening multiple unix sockets as long as the `local_socket_path` parameter does not conflict with an existing socket. 

When the `Proxy` object is created, it will start an "accept loop" async task which will open the unix server socket and listen for new connections. The `Proxy` should keep track of these async tasks so that they can be canceled when `Connector.close()` is called. 

When the "accept loop" task accepts a new connection, a new client socket is opened. Then the accept loop task will create a new async task to handle the client socket. This "client socket" task will do two things:

- call connector.connect() to establish a new socket to the server proxy
- create async tasks to copy data between the unix socket and the database instance socket

The client socket task will also need to handle the following error cases:
- When the `connector.connect()` fails, it should log an error and close the unix socket
- When the database socket is suddenly closed, we should close the unix socket
- When the unix client socket is suddenly closed, we should close the database socket

Cleanup: When connector.close() is called, the connector
   - cancel the "accept loop" task
   - close the unix socket server and delete the socket from the filesystem. 
   - briefly wait for all in-progress async tasks handling connections to close
   - close all open sockets.

I will added  additional comments in-line with the code.

Example of how to use Connector with psychopg and SqlAlchemy

```python
async def create_sqlalchemy_engine(
    instance_connection_name: str,
    user: str,
    password: str,
    db: str,
    ip_type: str = "public",
    refresh_strategy: str = "background",
    resolver: Union[type[DefaultResolver], type[DnsResolver]] = DefaultResolver,
) -> tuple[sqlalchemy.engine.Engine, Connector]:
    unix_socket_path="/tmp/conn"

    connector = Connector(refresh_strategy=refresh_strategy, resolver=resolver)
    await connector.start_unix_socket_proxy_async(
        instance_connection_name,
        user=user,
        password=password,
        db=db,
        local_socket_path=unix_socket_path,
        ip_type=ip_type,  # can be "public", "private" or "psc"
        autocommit=True,
    )

    # Construct the connection string
    # Note: The 'host' parameter points to the directory, not the socket file.
    connection_string = f"postgresql+psycopg2://{user}:{password}@{unix_socket_path}/{db}"
    
    # Create the SQLAlchemy engine
    engine = create_engine(connection_string)

    return engine, connector

```

Thank you again for your hard work on this. I look forward to seeing your next revision.
-Jonathan

hessjcg:
Hi @UzielSilva, I was able to spend some time on the tricky IO code to set up a local proxy server. This is still a work in progress, but I wanted to share it anyway. Please check out this branch: 

https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/tree/feat-local-proxy

Currently there are still some failing tests and edge cases that are not properly handled. Please feel free to continue my work.

UzielSilva:
Hi there, sorry for the lack of updates, took some mini vacations.
@hessjcg great stuff!!, I'm moving your proxy server into my branch. This is very helpful. :)

UzielSilva:
I'm having issues trying to fix one integration test, the psycopg one actually.

The issue is about a task that got Future attached to a different loop. The thing is that if I inject the running loop to the connector the issue appears to vanish but the connection now gives us a timeout error. All unit and integration tests are now passing, except for that one, and that is the only issue that is stopping me from finishing this PR. Can I get some help?

UzielSilva:
NVM, I solved it.

@hessjcg I think I'm ready for a new review.

seeker25:
This would be very useful using `pg8000` it's fairly slow especially when it comes to inserting or fetching many rows. I don't think `asyncpg` would be super easy to implement in an existing app that is synchronous. I believe `psycopg` would be much faster.

hessjcg:
Hi @UzielSilva Thank you for the updates! I apologize for the delay. I'll review this PR this week.

alberthli:
Agreed this feature would be really useful! Any updates?

JJZolper:
Yes, definitely hoping for updates soon. I have tried many paths but in Django it is a challenge due to really needing more TCP support. Some of the examples in other threads have given me ideas at least.

JJZolper:
I am definitely hoping to make use of this as soon as it is merged.

Jxio:
Looking forward to this being merged, this looks good would love to use it!

milos-blagojevic-tdly:
Same here, we are hoping to start using this asap!

sahibpandori:
+1, would love to use this once merged!

tbCortea:
any movement on this?

seeker25:
@hessjcg is it possible we could have this looked at please?

kgala2:
Thanks @UzielSilva, I've merged a https://github.com/GoogleCloudPlatform/cloud-sql-python-connector/pull/1452 as part of psycopg support in Python connector, psycopy should be available to use in Python connector's August release!

PR #1371, opened 2026-01-22
chore: Fix the build script so it works consistently with other projects

hessjcg:
This makes a number of small changes to the build.sh to make it work consistently with the other connector projects.


hessjcg:
This is out of date. Closing. The build script was fixed.

PR #1422, opened 2026-05-25
fix: update dependencies to latest

kgala2:
Updates project dependencies to their latest versions. Preserves the test extra tag for SQLAlchemy (SQLAlchemy[asyncio]) and fixes the broken nox session (removed the dangerous --freeze-outdated-packages flag that was corrupting requirements.txt).


noxfile.py
```diff
@@ -38,7 +38,7 @@ def deps(session):
         "pip-review",
     )
     session.run(
-        "pip-review","--auto","--freeze-outdated-packages",
+        "pip-review", "--auto",
```
panavenue, line 41: Curious - if this `--freeze-outdated-packages` intended?

panavenue (approved):

noxfile.py
hessjcg: --freeze-outdated-packages is not necessary. For some reason the "pip-review" tool uses the word "freeze" to mean "save the new version into requirements.txt."

hessjcg (approved):

PR #1476, opened 2026-09-11
fix: generate RSA keys off the event loop

andrew-kurin:
## Summary

`generate_keys()` is async, but it calls synchronous RSA generation and serialization directly. When a connector uses the application's event loop, key generation can delay other tasks on that loop. Moving the connector constructor to a thread does not address the key task scheduled on the supplied loop.

This change:

- Runs the existing generation/serialization body in `asyncio.to_thread`, without changing RSA parameters or PEM formats.
- Keeps the connector's key future on its existing loop. It does not move async database or HTTP clients between loops, change refresh strategy, or replace private key-cache state.
- Shields active generation from caller cancellation and drains it before propagating cancellation. Repeated cancellation does not release the waiter early; a simultaneous work failure is observed without replacing caller cancellation. Ordinary work failures still propagate.
- Adds tests for key parameters/formats and a matching signing/verification pair, loop progress during blocked work, ordinary failure, repeated cancellation with successful/failed work, and constructor loop/future ownership with two waiters.

## Testing

Run with `uv run --with nox nox --default-venv-backend uv -s lint unit-3.10 unit-3.12 unit-3.14`:

- `lint`: passed (Ruff, mypy, source-distribution build, and strict twine check).
- `unit-3.10`: 203 passed.
- `unit-3.12`: 203 passed.
- `unit-3.14`: 203 passed.
- `git diff --check`: passed.

Python 3.11/3.13 unit sessions and live Cloud SQL system/integration tests were not run locally. No cloud resources, credentials, or dependent application deployments were changed. This draft requests maintainer review of the cancellation contract; it is not a claim of live integration validation.


andrew-kurin:
@codex review

